PR #4, opened 2017-03-14
Unify and Automate maintenance of multiple logging targets

so0k:
This PR is to automate the creation of specific images for multiple logging targets.

Each target may have different `Dockerfile`, `fluent.conf` or `plugins` requirements. 
Plugins may be added as `gem install` instructions in the `Dockerfile`, or provided as 
ruby file in the respective `plugins` folder.

The aproach is copied from the `fluentd/fluentd-docker-image` repository, using
a  modified Makefile and Ruby templates.

Checklist:

- [x] Create directory structure for multiple logging targets
- [x] Create Dockerfile template for sample logging targets
- [x] Create fluent.conf template for sample logging targets
~- [ ] Create Tests to validate generated docker images~
~- [ ] Create Travis pipeline to run Tests~
- [x] Add Documentation on available tags and maintenance

- [ ] Create Automated Docker hub repo for this repository
  Somebody from treasure data should do this @repeatedly ?

I suggest to use the image name `fluent/fluentd-kubernetes`, Please advise.


so0k:
@tyranron I love your Makefile, please help review this PR
@edsiper as discussed yesterday.

so0k:
PS: I'm not using `-onbuild`, so I can remove the dummy `fluent.conf` that was at the root of the docker-image folder, I do however have to add the command to copy the plugins from the plugin folder.

This causes `.gitkeep` to be copied into the plugins folder as well, I haven't tested if this causes an error or not.

so0k:
If you would like to test the images built by this script, I have pushed them all here:
https://hub.docker.com/r/so0k/fluentd-kubernetes/tags/

c-knowles:
`.dockerignore` may be your friend for the gitkeep.

so0k:
I thought about adding `.dockerignore` but... I was worried about another make target & template to worry about... but I think this can just be a simple `copy everywhere` thing.

so0k:
Done, dockerignore added which will ignore `.gitkeep` at any level

so0k:
Added cloudwatch based on @icereval's work:

- https://github.com/kubernetes/charts/pull/211
- https://github.com/18F/cg-deploy-kubernetes/tree/master/images/fluentd-cloudwatch

so0k:
@c-knowles - what does the stackdriver setup look like?

so0k:
@repeatedly - can you please advise on what is missing to get this merged?

c-knowles:
@so0k looking good! Stackdriver setup is very similar to your work. It adds `fluent-plugin-google-cloud` gem as per the [gcp contrib image](https://github.com/kubernetes/contrib/tree/master/fluentd/fluentd-gcp-image) and then the outputs taken from [k8s addons](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/fluentd-gcp/fluentd-gcp-configmap.yaml#L312). Judging from the work that's gone on there recently it looks like stackdriver team will take ownership of that image.

My implementation is a sort of merge of the work there and here. For example they don't have the `kubernetes` custom parser. I'm running it outside of GCP with no stackdriver agent present so I also need to add credentials in. There are a few other useful bits contained there which we don't have here, either we could attempt to add those here or get the basics in and I can submit another PR once this one is done.

so0k:
I think a 2nd PR from you once this is merged is the best way forward.. may take a while though

I also wonder if it's ok to merge this without the automated testing at first..

so0k:
@frsyuki - need review and advise on how to get this merged

so0k:
@c-knowles - if you want to add your changes you can add it to https://github.com/honestbee/fluentd-kubernetes, I'll cherry pick to this branch 

I set up the fully automated docker build: https://hub.docker.com/r/honestbee/fluentd-kubernetes/

edsiper:
fyi: I will make sure to review this today

so0k:
any update?

c-knowles:
Based on comments in https://github.com/fluent/fluentd-kubernetes-daemonset/issues/6 I'm no longer sure there is a want to have Stackdriver/GCP logging added. The one thing I would add is that the official image is tailored a bit more towards those running on GCP whereas we're running on AWS. Plus they don't even add some things like the Kubernetes metadata filter which seems almost essential here. Maybe they will make it more agnostic but in the meantime I wouldn't mind adding support. @edsiper @so0k any particular thoughts?

so0k:
@c-knowles I'm not at all familiar with the stackdriver logging and if anyone else is using it while not on GCP, although it might be good for anyone if you are willing to add your work into this central location so anyone can built on top of it?

edsiper:
@c-knowles

My general perception when talking with different people from different organizations, is that Kubernetes metadata is a fundamental piece for them, not just Google.

Anyways I would like to hear more about more if the filter is generating some issues, so we can consider a way to disable it in the templates generation.

so0k:
Right now the metadata gem is in a shared gems folder so it's included for all builds and it shouldn't be hard to make exceptions to that rule

However I don't think that is what @c-knowles means as he said k8s metadata gem is essential to a fluent shipping container actually

edsiper:
you are right, I misunderstood.

c-knowles:
@edsniper, yes sorry if my comment was confusing. I meant the official image does not container the k8s metadata filter but from what I have seen it is pretty much essential. It is one reason to consider a stackdriver build here as well.

Sounds like we're good to add it then, will submit a PR.

Issue #11, opened 2017-04-11
Where is the image hosted?


gajus:
The links in the README.md take to 404 page.

https://hub.docker.com/r/fluent/fluentd-kubernetes/

gajus:
https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/

kenden:
Or even point to the docker store:
https://store.docker.com/community/images/fluent/fluentd-kubernetes-daemonset

alexandernilsson:
The DaemonSet yaml points to `quay.io/fluent/fluentd-kubernetes-daemonset`, which hasn't gotten updated in quite a while.

kenden:
https://store.docker.com/community/images/fluent/fluentd-kubernetes-daemonset/tags
Last updated a month ago, it's not that bad

kenden:
@alexandernilsson Good point, it's 3 months old. The version on dockerhub is 1 month old.
quay.io/fluent/fluentd-kubernetes-daemonset is not versioned, either, it's always 'latest'.

so0k:
Oops, seems I introduced that mistake as we use https://hub.docker.com/r/honestbee/fluentd-kubernetes/ instead of https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/ which is also what we use in our Helm charts.

edsiper:
Fixed,.

PR #17, opened 2017-05-19
update fluent-kubernetes-daemonset image location

superawesome:
The version on quay.io is old and not versioned, and seems to have problems with ES 5.x. This one on hub.docker.com is newer (and has tags for things other than just Elasticsearch) includes a much newer "elasticsearch" rubygem.

This should help with https://github.com/fluent/fluentd-kubernetes-daemonset/issues/11.


edsiper:
thanks!

PR #18, opened 2017-05-22
Add toleration to allow fluentd to run on masters

while1eq1:
Kubernetes 1.6 now has a taint on all the masters called "node-role.kubernetes.io/master" effect: NoSchedule, Which means without adding the toleration for this, Any DaemonSets will not run on masters. This PR adds this toleration to the fluentd daemonset.

### DIFF
--- a/fluentd-daemonset-elasticsearch.yaml
+++ b/fluentd-daemonset-elasticsearch.yaml
@@ -15,6 +15,9 @@ spec:
         version: v1
         kubernetes.io/cluster-service: "true"
     spec:
+      tolerations:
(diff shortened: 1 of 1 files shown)


edsiper:
Please correct me if I am wrong, but shouldn't be this a problem only when running in a single node cluster ?

apenney:
This is a problem for multi node clusters (if you want logging off pods on the master, which we do)

while1eq1:
Yes, like @apenney said, if you want logs from pods on masters (which we do as well)

edsiper:
thanks for the clarification.

PR #23, opened 2017-06-01
Use newer version of fluent-plugin-cloudwatch-logs

thiagophx:
By bumping the version we get rid of this [problem](https://github.com/ryotarai/fluent-plugin-cloudwatch-logs/pull/36).


edsiper:
thanks

PR #25, opened 2017-06-07
Add gcloud/stackdriver debian

c-knowles:
Alpine does not currently work due to issues with the grpc library on alpine. Even with `libc6-compat` installed, various errors still persist.


c-knowles:
@so0k got around to aligning our stack to this, unfortunately I couldn't immediately get the alpine stack to work due to issues like https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/issues/105, https://github.com/grpc/grpc/issues/8528. I tried installing `libc6-compat` but still hit various problems.

PR #26, opened 2017-06-07
Add docker to systemd sources

c-knowles:


edsiper:
What kind of logs does this retrieve ?

edsiper:
would you please solve the conflicts so we can rebase ?

PR #30, opened 2017-06-07
Ensure s3 plugin dir exists

chancez:


chancez:
oops wait, need to regenerate

PR #36, opened 2017-06-20
v0.12/alpine-s3: Add missing 'fluent-plugin-systemd' plug-in

fhemberger:
Took the lines from the Debian counterpart.


edsiper:
thanks for catching this. Note that the template must be updated instead of the generated Docker files. Check the Makefile and it rules to regenerate everything.

PR #41, opened 2017-07-12
Revert "Mmake ES user/password inclusion conditional"

dileepfrog:
Reverts fluent/fluentd-kubernetes-daemonset#40


dileepfrog:
This ERB template gets interpolated by the Makefile that builds the individual images and will inherit its environment. This means that FLUENT_ELASTICSEARCH_USER and FLUENT_ELASTICSEARCH_PASSWORD will always be blank, removing them from all generated fluent.conf which may break non-AWS ES use cases.

The right fix to this is to perform the interpolation and create fluent.conf from within Docker image itself as part of the bootstrap process when the env vars relevant are actually present.

edsiper:
just merged, I am regenerating the templates

Issue #57, opened 2017-10-12
Can't add multiple tags to Loggly logs


ZachEddy:
Hi there,

I'd like to add multiple tags to logs that are sent from Fluentd to Loggly. With the current Fluentd configuration, Loggly logs are only tagged as `fluentd` and I don't have the option to add my own tags.

It would be handy to pass a collection of tags to`fluentd.conf.erb` as an environment variable via my Kubernetes YAML file — maybe something like this:
```erb
<%# -- fluentd.conf.erb -- %>

<%# current Loggly configuration %>
loggly_url "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/fluentd"
<%# proposed Loggly configuration %>
loggly_url "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/#{ENV['LOGGLY_TAGS'] || 'fluentd'}"
```

This seems like a useful feature and I'm happy to set to work on a pull request. What do other people think?

If you're not familiar with Loggly, [here's](https://www.loggly.com/docs/tags/) some info about tagging.

cosmo0920:
This should be merged. Closing.

PR #58, opened 2017-10-12
Add support for custom Loggly tags

ZachEddy:
**Custom Loggly Tags**

This pull request allows Loggly users to customize how their logs are tagged. Inside issue #57, I describe why this functionality might be useful:

> I'd like to add multiple tags to logs that are sent from Fluentd to Loggly. With the current Fluentd configuration, Loggly logs are only tagged as `fleuntd` and I don't have the option to add my own tags.

The proposed change enables someone to provide Fluentd with a list of custom tags. Fluentd will tag all Loggly logs according to the list of given tags. Alternatively, if no custom tags are provided, all logs are tagged as `fluentd` by default (to reflect the current behavior).

### DIFF
--- a/templates/conf/fluent.conf.erb
+++ b/templates/conf/fluent.conf.erb
@@ -41,7 +41,7 @@
 <match **>
    type loggly
    log_level info
-   loggly_url "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/fluentd"
+   loggly_url "https://logs-01.loggly.com/inputs/#{ENV['LOGGLY_TOKEN']}/tag/#{ENV['LOGGLY_TAGS'] || 'fluentd'}"
 </match>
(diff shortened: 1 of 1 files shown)


pirankashani:
Hello,
I would like to customize my log with more tags. for instance if a log look likes as following:

> {"log":"I1125 11:33:18.947057       1 trace.go:76] Trace[1459855838]: \"Get /api/v1/namespaces/kube-system/endpoints/kube-scheduler\" (started: 2017-11-25 11:33:18.315642294 +0000 UTC) (total time:  631.38115ms):\n","stream":"stderr","time":"2017-11-25T11:33:18.952213987Z"}

I need to arrange it in this way:
`I       = Info`
`1125= 25-Dec`
and other items so on.

Is it possible to edit this project anyhow or I should create a Fork request? What is your idea?

PR #75, opened 2018-02-16
Fix shebang as tests won't work with sh

bliemli:
Executing `/fluentd/entrypoint.sh` throws the following errors because `sh` doesn't know the '[[' construct:

```
/fluentd/entrypoint.sh: 5: /fluentd/entrypoint.sh: [[: not found
/fluentd/entrypoint.sh: 9: /fluentd/entrypoint.sh: [[: not found
```
Changing the shebang to `bash` fixes this.

### DIFF
--- a/docker-image/v0.12/alpine-cloudwatch/entrypoint.sh
+++ b/docker-image/v0.12/alpine-cloudwatch/entrypoint.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -e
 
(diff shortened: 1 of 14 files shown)


errm:
Alpine doesn't ship with bash...

I think it would be better to switch from:
```
if [[ -z ${FLUENT_ELASTICSEARCH_USER} ]] ; then
```
to
```
if [ -z "$FLUENT_ELASTICSEARCH_USER" ] ; then
```

edsiper:
thanks for the feedback on this. entrypoint fixed in master with simple [ ]

PR #78, opened 2018-02-22
Add exception detection plugin to elastic version

vaivars:
Elastic works better if stacktrace errors are passed together instead of having each line separately.


carlosjgp:
Duplicates #72

edsiper:
Fixed by 7949a2c4f6df91924c8ed79dbed2ee3b88091da0

Issue #69, opened 2018-01-31
Elasticsearch images install fluentd-1.1.0


Stelminator:
At least on debian.

```
# gem install fluent-plugin-elasticsearch
Building native extensions.  This could take a while...
Successfully installed strptime-0.2.3
Fetching: dig_rb-1.0.1.gem (100%)
Successfully installed dig_rb-1.0.1
Fetching: fluentd-1.1.0.gem (100%)
[...]
```

Stelminator:
found the source https://github.com/uken/fluent-plugin-elasticsearch/commit/162c69e86e8bec17f9873e60c317598c18c9b72f

Stelminator:
going back farther: https://github.com/uken/fluent-plugin-elasticsearch/commit/ca48c1d9d86886b0f32a273c330175d6f22558ef

I think this might come from what gem version got published, rather than any recent source changes.

Stelminator:
yup, [2.0.0](https://rubygems.org/gems/fluent-plugin-elasticsearch/versions/2.0.0) got released on November 06, 2017, and the [last build before that](https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/builds/bjajeafhrzsbrzhasszuv89/) happened on October 02.

Issue #43, opened 2017-07-12
Re-design how Fluentd should be deployed as a Daemonset


edsiper:
Current structure of this suggested way to deploy Fluentd as a daemonset is generating some complications to maintain different based distro, plugins and different setups. 

Taking in count @chancez suggestions, I think this could work as:

- One debian base image:
  -  Deprecate Alpine because of:  cannot work with systemd, hard to maintain with Jemalloc enabled (required to reduce memory usage because of high glibc memory fragmentation).
  - Debian image by Google  (gcr.io/google-containers/debian-base-amd64:0.1) is really secure, 
- Use ConfigMaps instead of configuration files inside the images
- Similar documentation to the one existing by [CoreOS/Tectonic: Logging](https://coreos.com/tectonic/docs/latest/admin/logging.html)

Please share your thoughts and ideas.

chancez:
One of the biggest pain points is plugins. I'd like to experiment with a few ideas, when I find some more time, but I'll put them here so others can feel free to try it if they like:

1) Create a base image with most plugins people might need, not including most outputs (install fluent-plugin-rewrite-tag-filter, fluent-plugin-prometheus, fluent-plugin-kubernetes_metadata_filter, fluent-plugin-rewrite-tag-filter, the forwarder plugins, maybe a few more).
2) Creating an image based on the base image with most output plugins pre-installed. This image will be necessarily larger, but it will be an "all-in-one" for most people.
3) Creating an image based on the base image which can dynamically install plugin gems based on a list of gems in an environment variable, at startup.

I'm not a huge fan of option 3, but I am curious to see how well it works in practice since the pods only need to install the gems the first time they're started, restarts wouldn't need to re-download the gem, only if the pod is re-created does it need to re-download, so this may work for some use cases.

By building a base image containing everything except output plugins specific to usage with Kubernetes, it's really easy to extend the image with your own modifications, for your specific logging destination, and if you don't care about the size as much, the all-in-one image would have most output plugins people need already installed.

PR #86, opened 2018-03-19
Use Bundler to manage gem dependencies

### LINKED ISSUE #27: Install via Gemfile?
Is there any specific reason we wish to put the `gem install`s directly in the Dockerfile or could we [use a Gemfile](https://github.com/kubernetes/contrib/blob/master/fluentd/fluentd-gcp-image/Dockerfile#L38) for each permutation?

errm:
* Makes builds more deterministic
* Avoids issues with plugin incompatibility with fluentd version
* Vulnerability scanning from github
* ~ moves a little closer to solving #43 perhaps
* Fixes #27


errm:
Also fixes #69

edsiper:
@repeatedly any comments on using Bundler instead of gem utility ?

repeatedly:
Gemfile based plugin management is easier than list of install command.
So the approach seems good for me.

errm:
I am working on rebasing this PR on the recent changes ...

errm:
Rebase seems good... I had to add git to the build process to cover the gems that are now being installed from git repos. @edsiper could you take another look?

repeatedly:
@edsiper Do you have any concern? We need to release v1 based images so need to merge this change before.

repeatedly:
@edsiper Can we merge?

repeatedly:
No counter opinion. Merged. Thanks!

Issue #87, opened 2018-03-19
RBAC policy example required


jpds:
Hello,

I tried deploying `fluent/fluentd-kubernetes-daemonset:v0.12-alpine-s3` into an RBAC enabled Kubernetes cluster and received this error:

```
2018-03-19 17:51:33 +0000 [error]: config error file="/fluentd/etc/fluent.conf" error="Exception encountered fetching metadata from Kubernetes API endpoint: pods is forbidden: User \"system:serviceaccount:kube-system:default\" cannot list pods at the cluster scope ({\"kind\":\"Status\",\"apiVersion\":\"v1\",\"metadata\":{},\"status\":\"Failure\",\"message\":\"pods is forbidden: User \\\"system:serviceaccount:kube-system:default\\\" cannot list pods at the cluster scope\",\"reason\":\"Forbidden\",\"details\":{\"kind\":\"pods\"},\"code\":403}\n)"
````

**Comment by medanasslim** [comment]
I have the same issue. Anyone has a solution ?

**Comment by rbtr** [comment]
```
--- 
apiVersion: v1 
kind: ServiceAccount 
metadata: 
  labels: 
    k8s-app: fluentd
  name: fluentd
  namespace: kube-system 
 
--- 
kind: ClusterRole 
apiVersion: rbac.authorization.k8s.io/v1beta1 
metadata: 
  name: fluentd-clusterrole 
rules: 
  - apiGroups: 
      - "" 
    resources: 
      - "namespaces" 
      - "pods" 
    verbs: 
      - "list"
      - "get" 
      - "watch" 
 
--- 
kind: ClusterRoleBinding 
apiVersion: rbac.authorization.k8s.io/v1beta1 
metadata: 
  name: fluentd-clusterrole 
roleRef: 
  apiGroup: rbac.authorization.k8s.io 
  kind: ClusterRole 
  name: fluentd-clusterrole 
subjects: 
- kind: ServiceAccount 
  name: fluentd
  namespace: kube-system
```

**Comment by lucj** [comment]
Hi, I have the following error when creating the clusterrole. Any idea what I'm missing ?

$ kubectl create -f clusterrole.yaml


```
  Error from server (Forbidden): error when creating "test-clusterrole.yaml": 
  clusterroles.rbac.authorization.k8s.io "fluentd-clusterrole" is forbidden: attempt to grant extra privileges: 
[...]

PR #89, opened 2018-03-28
Fixed RBAC policy example required -- Added rbac full example

ChrsMark:
Fixed https://github.com/fluent/fluentd-kubernetes-daemonset/issues/87.

cc: @edsiper could you please take look and say if this could be useful to be included?

### DIFF
--- a/fluentd-daemonset-elasticsearch-rbac.yaml
+++ b/fluentd-daemonset-elasticsearch-rbac.yaml
@@ -1,0 +1,96 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: fluentd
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks! My friend confirmed this patch works on k8s 1.7 and 1.9.

aruruka:
I'm using a shared k8s cluster, and the admin won't allow me to query pods at the cluster scope.
Is there any solution in this case?
I noticed this problem at [Restricting API calls to a single namespace](https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/issues/88). It seems that fluentd always queries at cluster scope?

PR #100, opened 2018-04-17
Fix missing Graylog tag and add Graylog RBAC example

wsimmonds:
This commit features a minor change - adding a missing </match> tag to the Graylog configuration.

It also has an example Graylog DaemonSet configuration with RBAC enabled. Unlike "elasticsearch-rbac" it does not include:

- kubernetes.io/cluster-service: "true"

Whilst testing this on a GKE cluster, the inclusion of this resulted in it always being killed off after under a minute with the message "Killing container with id docker://fluentd:Need to kill Pod"


errm:
Looks like this was fixed by #144

repeatedly:
Thanks for the patch but other patch seems to include this fix.

Issue #98, opened 2018-04-16
Logz.io image not being published


LudwigTirazona:
I was looking through the repo, and while I see come configs for Logz.io, it seems that the image itself is not being built and published.

Is that an oversight, or is it intentionally being left unbuilt?

hobbsh:
Running into this as well

PR #102, opened 2018-04-17
Add debian build to logz.io

jamielennox:
There is a debian-logzio folder from which we are supposed to build an
image however it's not part of the image list and so is not being
updated or built.

Add this to the Makefile.


jamielennox:
Screen Shot

jamielennox:
Note that this doesn't solve #98 - does that just require another run?

repeatedly:
You need to run `make src-all`.

jamielennox:
looks fixed by #118

PR #105, opened 2018-04-20
Added fluent-plugin-rewrite-tag-filter

blaketastic2:
We use this to strip off the "kubernetes.var.log.containers." prefix


blaketastic2:
I'm gonna re-do this...

PR #107, opened 2018-04-24
Added the rewrite tag plugin

blaketastic2:
Added the ability to re-write tags


repeatedly:
rewrite-tag-filter is widely used so adding it seems useful. Thanks!

PR #110, opened 2018-04-25
added aws key/secret fields in fluentd config

V3ckt0r:
Signed-off-by: V3ckt0r <burhan_abdi@hotmail.co.uk>

### DIFF
--- a/docker-image/v0.12/alpine-s3/conf/fluent.conf
+++ b/docker-image/v0.12/alpine-s3/conf/fluent.conf
@@ -6,11 +6,13 @@
 
 <match **>
   # docs: https://docs.fluentd.org/v0.12/articles/out_s3
-  # note: this configuration relies on the nodes have an IAM instance profile with access to your S3 bucket
+  # note: Make sure you use secrets to store your AWS key/secret
   type s3
   log_level info
   s3_bucket "#{ENV['S3_BUCKET_NAME']}"
   s3_region "#{ENV['S3_BUCKET_REGION']}"
+  aws_key_id "#{ENV['AWS_KEY_ID']}"
+  aws_sec_key "#{ENV['AWS_SEC_KEY']}"
   s3_object_key_format %{path}%{time_slice}/cluster-log-%{index}.%{file_extension}
   time_slice_format %Y/%m/%d
   time_slice_wait 10m
--- a/docker-image/v0.12/debian-s3/conf/fluent.conf
+++ b/docker-image/v0.12/debian-s3/conf/fluent.conf
@@ -7,11 +7,13 @@
 
 <match **>
   # docs: https://docs.fluentd.org/v0.12/articles/out_s3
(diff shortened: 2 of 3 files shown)


V3ckt0r:
We shouldn't be relying on IAM roles on the EC2 host to filter down to the container. With RBAC this make it tricky. It seems much simpler to simply set up the key and secret as secrets in k8s then pull them into the daemonset like so:

```
kind: DaemonSet
metadata:
  name: fluentd
  namespace: default
  labels:
    app: fluentd
spec:
  selector:
    matchLabels:
      name: fluentd-s3
  template:
    metadata:
      labels:
        name: fluentd-s3
    spec:
      containers:
      - name: fluentd
        image: vect0r/fluentd-kubernetes-daemonset:latest
        env:
          - name: S3_BUCKET_NAME
            value: "test1"
          - name: S3_BUCKET_REGION
            value: "eu-central-1"
          - name: AWS_KEY_ID
            valueFrom:
              secretKeyRef:
                name: secret-s3
                key: secret_access_s3
          - name: AWS_SEC_KEY
            valueFrom:
              secretKeyRef:
                name: secret-s3
                key: secret_key_s3
        ports:
          - containerPort: 24220
      serviceAccountName: fluentd
```

repeatedly:
Does this break existing environment, right?
fluent-plugin-s3 prefers key/secret setting so this change causes empty key/secret setting and it seems to cause credential error.

V3ckt0r:
How do you mean by break? If the key/secret is not provided, like above then yes the pod will fail to start up. But I think that is better than depending on the pod running on a node that as IAM access to the s3 bucket. Due to the issue of all other pods running on that node also having access.

I've tested these changes in my k8s cluster with the above daemonset, fluentd comes up successfully. I've created an image here you can use to take a look and test https://hub.docker.com/r/vect0r/fluentd-kubernetes-daemonset/.

repeatedly:
If someone wants to use node's IAM credential, how to specify it?

V3ckt0r:
Hey @repeatedly,

So what I'm thinking is if we are just relying on IAM to provide the S3 access, then we have a security problem as we would be giving pods running on the same node access to s3 too. To get around this you will need to deploy something like https://github.com/jtblin/kube2iam in your cluster to block the IAM role to just the specific pods that need it.

If users don't do this they can potentially have a security issues later down the line without realising. I think the default should be changed to explicitly use key/secret as this can avoid the issue without having to deploy something like aws2iam. (by not relying on a blanket IAM role on the node). 

What are your thoughts on this?

errm:
I don't think this is OK to merge... It will break the setup of anyone who is relying on this configuration when using IAM profiles ...

errm:
I think best practice would be to use kube2iam or kaim to provide a useable IAM role ... I think this change is not needed...

icecold21:
Plus one for this actually

PR #115, opened 2018-04-26
Use Loggly's bulk endpoint

hobbsh:
Not sure how common this is but I was getting a lot of "Error connecting to loggly please check your url"

Following the KBs noted in [this forum post](https://community.loggly.com/customer/portal/questions/16719622-getting-error-connecting-to-loggly-verify-the-url) switching to the bulk endpoint has made these errors appear much less frequently. For further reading, see this [Loggly KB](https://www.loggly.com/docs/http-bulk-endpoint/)

### DIFF
--- a/templates/conf/fluent.conf.erb
+++ b/templates/conf/fluent.conf.erb
@@ -45,7 +45,7 @@
    @type loggly
    @id out_loggly
    log_level info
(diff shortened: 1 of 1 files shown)


repeatedly:
I can't read forum post because I don't have an account but Loggly KB uses bulk so this patch seems no problem.

PR #117, opened 2018-04-27
fluentd-cloudwatch image has updated dependencies

jpke:
not sure how to build this into the module's automation. but needed to update gems installed in the fluentd-cloudwatch image in order to parse json logs. this fixed https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs/issues/65 for me.

also needed for the fix was fluend-config update to include:
```  
  <filter kubernetes.**>
    @type kubernetes_metadata
    @id filter_kube_metadata
    char_encoding utf-8
  </filter>

  <match **>
    @type cloudwatch_logs
    log_group_name "#{ENV['LOG_GROUP_NAME']}"
    auto_create_stream true
    use_tag_as_stream true
    json_handler yajl
  </match>
```


repeatedly:
fluent-plugin-cloudwatchlogs v0.5 doesn't work with fluentd v0.12 so don't update dependency of fluentd v0.12.
Try use v1.x images for it.

PR #118, opened 2018-04-29
Add debian logzio

### LINKED ISSUE #98: Logz.io image not being published

repeatedly:
Fix #98 
See https://github.com/fluent/fluentd-kubernetes-daemonset/pull/102


hobbsh:
thank you @repeatedly 🍾

PR #119, opened 2018-05-02
Fix fluent user issue

repeatedly:
To fix permission issue with `fluent` user.
Use micdah's approach.
Adding group by monotek is impossible because parent image creates `fluent` user in `entrypoint.sh`.


micdah:
The new '[Entrypoint]` defined pointing to `/fluentd/entrypoint.sh` doesn't spawn using `dumb-init` but rather spawns directly using `/bin/sh`, I guess that might be a potential issue in regards to signal handling and the other stuff which _dumb-init_ handles (https://github.com/Yelp/dumb-init)

Just my 2 cents.

repeatedly:
Ah, I forgot that point. So we should change following line to `#!/usr/bin/dumb-init /bin/sh`?

https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/templates/entrypoint.sh#L1

micdah:
@repeatedly Yes I believe that should work, pretty much the setup I am using currently to run it which seems to work as intended. 😄

PR #129, opened 2018-05-10
Remove redundant CMD from v1.1 images

errm:
Signed-off-by: Ed Robinson <edward-robinson@cookpad.com>


repeatedly:
Ah, I forgot to remove it. Thanks!

PR #132, opened 2018-05-11
Fixes the links in the README

errm:
For some reason they stopped working after my last change, even though
there was no diff... I can't see how they ever worked though


repeatedly:
Thanks!

PR #146, opened 2018-06-13
Bump fluent-plugin-systemd

errm:
Pulls in version 1.0.1 or 0.0.11  (for v0.12) of fluent-plugin-systemd

See https://github.com/reevoo/fluent-plugin-systemd/pull/65 for more
info about the issue that it fixes.

This may manifest in journal files not being correctly deleted by
some versions of systemd after they have been rotated.

FYI there are some other plugin updates that show up when running `make gemfile-all`... I did not include them in this PR...


repeatedly:
I understood the background. Thanks!

PR #148, opened 2018-06-13
install gem bundler for debian images

### LINKED ISSUE #147: gem bundler missing from debian images
Using image `v1.2-debian-elasticsearch` trying to include custom Gem specs via Gemfile which is mounted in at run time.

Set `FLUENTD_DISABLE_BUNDLER_INJECTION=0` so that the runtime bundler is executed.

Container fails to start because gem bundler is not installed.
```
error: Could not find 'bundler' (>= 0) among 49 total gem(s)
Checked in 'GEM_PATH=/fluentd/vendor/bundle/ruby/2.3.0', execute `gem env` for more information
```

Fix: `gem install bundler`

therealgambo:
Fixes #147 

@repeatedly LGTM

### DIFF
--- a/docker-image/v0.12/debian-cloudwatch/Dockerfile
+++ b/docker-image/v0.12/debian-cloudwatch/Dockerfile
@@ -19,7 +19,7 @@ COPY Gemfile* /fluentd/
      && apt-get install \
      -y --no-install-recommends \
      $buildDeps \
-    && gem install bundler --version 1.16.1 \
+    && gem install bundler --version 1.16.2 \
     && bundle config silence_root_warning true \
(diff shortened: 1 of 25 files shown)


errm:
You should not need to install bundler via the system package `ruby-bundler` and with rubygems ... if we need it to be installed with rubygems please remove the system package...

therealgambo:
Removed system package in favour of rubygems.

@errm lgtm

therealgambo:
Done.

errm:
Nice LGTM

repeatedly:
Thanks. Just merged.

@errm Thanks for the reviews 👍

PR #155, opened 2018-06-25
Include rewrite-tag-filter plugin for fluentd

naiduarvind:
Signed-off-by: naiduarvind <techmaxed.net@gmail.com>


naiduarvind:
I created a new clean PR that passes the checks. If you could review and let me know if there is anything else that needs to be fixed.

cc: @errm @repeatedly

templates/Gemfile.erb
```diff
@@ -12,6 +12,9 @@ gem "oj", "3.5.1"
 <% if !is_v1 %>
 gem "fluent-plugin-secure-forward"
 gem "fluent-plugin-record-reformer"
+gem "fluent-plugin-rewrite-tag-filter", "<2.0.0"
+<% else %>
```
errm: I don't think we need these version constraints here, the version constraints seem fine on the gem ...

errm: Also it might be nicer to bump the versions up for all the other gems in a separate commit from adding `fluent-plugin-rewrite-tag-filter`, it will make the history cleaner ...

naiduarvind: > Also it might be nicer to bump the versions up for all the other gems in a separate commit from adding fluent-plugin-rewrite-tag-filter, it will make the history cleaner ...

repeatedly: With bundler, correct version is installed because bundler uses different algorithm for resolving version dependency.

You can check the behaviour with following Gemfile.

```
source 'https://rubygems.org/'

#gem 'fluentd', '~> 1' # install rewrite-tag-filter v2
gem 'fluentd', '~> 0.12.0' # install rewrite-tag-filter v1
gem 'fluent-plugin-rewrite-tag-filter'
```

naiduarvind:
I have removed the version constraints. Do have a look at the generated files. Is there anything else that needs to be done for this PR to be merged?

cc: @errm @repeatedly

repeatedly:
Thanks!

PR #160, opened 2018-07-02
Replace deprecated MAINTAINER tag with LABEL

jgunnink:
Per documentation on https://docs.docker.com/engine/reference/builder/#maintainer-deprecated using MAINTAINER in dockerfiles is now deprecated.

Replaced affected files with LABEL syntax.


repeatedly:
Looks good. Thanks!

PR #178, opened 2018-07-31
Fix proposal for pos_file issue in /var/log

mbolek:
This should allow for configurable pos_file directory based on FLUENT_POS_DIR ENV variable


repeatedly:
Does this work on your environment?
This patch changes pos_file directory at image building time, not fluentd runtime.
If you want to change configuration parameter at runtime, need to use this feature: https://docs.fluentd.org/v1.0/articles/config-file#embedded-ruby-code

repeatedly:
BTW, please add your sign to commits: https://github.com/apps/dco

mbolek:
I'm using configmap to change fluentd's config file so I've changed the pos location and it seems to run but the thing is a regular user (fluent with UID 1000) cannot access files from /var/log/containers which ends up with error, so I guess it's not a proper direction as it still needs root-level access to those files :(

Second thing being the config file itself where we'd have to validate the ENV variable within each pos_file on runtime if I understand correctly?
so it would be like this:
```
pos_file "#{ENV['FLUENT_POS_DIR'] != nil ? ENV['FLUENT_POS_DIR'] : '/var/log'}/fluentd-kube-controller-manager.log.pos"
```
It works when added to the configmap, but the access issue remains

repeatedly:
> Second thing being the config file itself where we'd have to validate the ENV variable within each pos_file on runtime if I understand correctly?

Yes. Your patch approach doesn't work because erb is evaluated at docker image building phase.
So fluentd uses fixed value, in this case `/var/log`, in user environment.

`pos_file "#{ENV['FLUENT_POS_DIR'] != nil ? ENV['FLUENT_POS_DIR'] : '/var/log'}/fluentd-kube-controller-manager.log.pos"`

This approach works because this line is evaluated at fluentd startup phase.

PR #184, opened 2018-08-24
add Azure Eventhubs image target

sputnik13:


sannadi99:
@sputnik13 Is this v1.2-debian-azureeventhubs is available in fluentd-kubernetes-daemonset for use. We are using fluentd daemonset for collecting logs and trying to find out a plugin to send those logs to azure eventhubs. It seems you tried the same here. So just thought to check with you for some help.

PR #185, opened 2018-08-29
Use type_name parameter explicitly

cosmo0920:
Compiatibility with Elasticsearch 6.x.

Multiple mapping types are not supported in Elasticsearch 6.x.  By
default [fluent-plugin-elasticsearch](https://github.com/uken/fluent-plugin-elasticsearch) creates two mapping types per index `[foo,fluentd]`.
See [issue](https://github.com/uken/fluent-plugin-elasticsearch/issues/442)


repeatedly:
Thx!

PR #97, opened 2018-04-13
Added Prometheus Metrics.

kousik93:
Added fluent-plugin-prometheus exposing metrics on the default 24231 port. 
Issue: https://github.com/fluent/fluentd-kubernetes-daemonset/issues/93

Signed-off-by: Kousik Sundara Padmanabhan <kpadmanabhan@proofpoint.com>

Issue #93, opened 2018-04-06
Add Prometheus metrics support


jpds:
https://docs.fluentd.org/v0.12/articles/monitoring-prometheus states that Prometheus should be used in conjunction with Kubernetes+fluentd however the Docker images from this repo do not include any support for Prometheus at all.

Can this please be added and annotations added to the daemonset?

KYannick:
@max-rocket-internet you added prometheus, I'm currently using v1.8.1-debian-cloudwatch-1.0 of the image, should this be enabled by default? I'm not passing parameters to disable the metrics, but I'm not seeing metrics at the moment.

max-rocket-internet:
@KYannick this was a while ago and I no longer use fluentd so can't quite remember but looking that the PR, in your container, do you have `/fluentd/etc/prometheus.conf`?

PR #198, opened 2018-09-20
Prometheus metrics added

gitirabassi:
Closing #93 and #97 

Added https://github.com/fluent/fluent-plugin-prometheus by default on port 24231


schndr:
Howdy, any support needed? lgtm

max-rocket-internet:
@gitirabassi can you rebase to fix the conflicts?

@repeatedly can we get this reviewed? A lot of people are interested in prometheus support 😃

max-rocket-internet:
@repeatedly there have also been 2 other PRs to try and address this: https://github.com/fluent/fluentd-kubernetes-daemonset/pull/96 https://github.com/fluent/fluentd-kubernetes-daemonset/pull/97

repeatedly:
Currently, v0.12 images are outedated so no need to apply changes to v0.12 images.
In addition, this PR has only `prometheus` input plugin. Is it enough?

max-rocket-internet:
@repeatedly it seems this PR might be dead. I will make a new one. A question though: is it possible to enable/disable plugins based on an env var? Because it would be nice to have prometheus support optional but easy to enable.

repeatedly:
@max-rocket-internet Currently, no envver based if in configuration syntax but we can use `@include` approach to emulate it.

https://github.com/fluent/fluentd-kubernetes-daemonset#disable-systemd-input
https://github.com/fluent/fluentd-kubernetes-daemonset/blob/6fa634a8dd02daf826b5a914fddd8b2cbed5aba4/templates/conf/fluent.conf.erb#L8

`@include` ignores error when target file is missing.

max-rocket-internet:
OK good idea @repeatedly, thanks! See here: https://github.com/fluent/fluentd-kubernetes-daemonset/pull/287

Issue #154, opened 2018-06-21
queue size exceeds limit


rainard:
2018-06-21 08:51:11 +0000 [warn]: emit transaction failed: error_class=Fluent::BufferQueueLimitError error="queue size exceeds limit" tag="kubernetes.var.log.containers.calico-node-qr9gm_kube-system_calico-node-70d2ef3123bf34b4f35a77e7d1bd3ea1c212989475f8d70bc08e17846dd6ca63.log"

How to solve the problem?

PR #202, opened 2018-09-27
Made fluentd buffer params configurable

goruha:
## What
* Allow to config buffer params with env vars

## Why
* Fix `BufferQueueLimitError`

Related:
https://github.com/fluent/fluentd-kubernetes-daemonset/issues/154


repeatedly:
Merged.

PR #211, opened 2018-10-24
Use fluent-plugin-remote_syslog for Fluentd v1.2

okkez:
Replace fluent-plugin-kubernetes_remote_syslog with fluent-plugin-remote_syslog.
Because fluent-plugin-kubernetes_remote_syslog uses old plugin API and
fluent-plugin-remote_syslog is well maintained.


repeatedly:
New PR includes this change and it is merged. Thanks!

PR #224, opened 2018-11-07
Add example logentries daemonset

### LINKED ISSUE #37: Better document Logentries usage?
I'm not sure how to use this with Logentries. As far as I undertand I would need to mount a configmap at `/etc/logentries/tokens.yaml` but I'm not sure how this file should look like.

medzin:
Add example logentries daemonset. Resolves #37.


repeatedly:
Thanks!

PR #227, opened 2018-11-20
ship logs to loomsystems

yatirdagan:


repeatedly:
Do other users want this image?
I'm not sure loomsystems is popular in k8s environemnt so
I can't judge loomsystems destination should be supported or not in official images.

yatirdagan:
Hi,

Loom Systems is used by many customers, including several fortune-500
companies.
We have implemented this PR after being requested by several of our
customers.

We're happy to provide whatever information you need to decide in favor of
including us.

Thanks

‫בתאריך יום ג׳, 4 בדצמ׳ 2018 ב-16:53 מאת ‪Masahiro Nakagawa‬‏ <‪
notifications@github.com‬‏>:‬

fluent-kubernetes-loomsystems.yaml
```diff
@@ -0,0 +1,55 @@
...
+        effect: NoSchedule
+      containers:
+      - name: fluentd
+        image: loomsystems/fluentd-kubernetes-loomsystems:v0.12.43-loomsystems
+        imagePullPolicy: Always
+        env:
+          - name:  FLUENT_UID
+            value: "0"
+          - name:  LOOM_RECIEVER_HOST
+            value: "129.213.38.169"
```
DavidTPate: Is this the correct IP? It doesn't resolve to any domain name.

fluent-kubernetes-loomsystems.yaml
```diff
@@ -0,0 +1,55 @@
...
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      containers:
+      - name: fluentd
+        image: loomsystems/fluentd-kubernetes-loomsystems:v0.12.43-loomsystems
+        imagePullPolicy: Always
+        env:
+          - name:  FLUENT_UID
+            value: "0"
```
DavidTPate: In the other images we don't set this by default but leave it to the user. Don't think we should be changing that here.

fluent-kubernetes-loomsystems.yaml
```diff
@@ -0,0 +1,55 @@
...
+          - name:  LOOM_RECIEVER_HOST
+            value: "129.213.38.169"
+          - name:  LOOM_RECIEVER_SSL_PORT
+            value: "9999"
```
DavidTPate: Are these ports correct? They are non-standard ports.

fluent-kubernetes-loomsystems.yaml
yatirdagan: fixed

fluent-kubernetes-loomsystems.yaml
yatirdagan: fixed

fluent-kubernetes-loomsystems.yaml
yatirdagan: fixed, these setting will be defined by the user

fluent-kubernetes-loomsystems.yaml
```diff
@@ -0,0 +1,54 @@
...
+      tolerations:
+      - key: node-role.kubernetes.io/master
+        effect: NoSchedule
+      containers:
+      - name: fluentd
+        image: loomsystems/fluentd-kubernetes-loomsystems:v0.12.43-loomsystems
+        imagePullPolicy: Always
+        env:
+        - name:  LOOM_RECIEVER_HOST
+          value: "{{RECEIVER HOST}}"
```
DavidTPate: If these are meant to be set by the user we should mention in the docs that they are expected to set these.

repeatedly:
BTW, we dropped v0.12 and alpine version support before: https://github.com/fluent/fluentd-kubernetes-daemonset#alpine-linux-this-is-deprecated-use-debian-images-instead
If you want to support loomsystems, please use debian images.

And you need to add you to maintainer list: https://github.com/fluent/fluentd-kubernetes-daemonset#maintainers

kenhys:
Hi, we can't afford to support suggested image
because of limitation about max size of build pipeline on
hub.docker.com for a long time.
This is why this PR was marked pending.

Recently, the above situation was resolved by migrating to GitHub Actions.

If you still interested in supporting this feature in fluent-kubernetes-daemonset,
Could you mind to create a new PR, please?

Thank you for contributing this project!

PR #232, opened 2018-11-28
Give the option to configure the path on elasticsearch

vigohe:


repeatedly:
@cosmo0920 Does empty string valid for `path` parameter? By default, `ENV['FLUENT_ELASTICSEARCH_PATH']` is nil so it becomes empty string in `configure`.

cosmo0920:
Es plugin just pass through `path` parameter and other connection parameters into `elasticsearch-ruby`.
It seems that empty string is valid for `path`.
Because it is used in string interpolation: https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/transport/base.rb#L238

repeatedly:
@cosmo0920 Thanks.

@vigohe If you want to apply this change to images, please run `make src-all`.

repeatedly:
Sorry for the late. Just merged.

vigohe:
Thanks 💪

Issue #181, opened 2018-08-11
json log not getting parsed to the output record fields


ericln:
I have tried using both `fluent/fluentd-kubernetes-daemonset:v0.12-alpine-elasticsearch` and `fluent/fluentd-kubernetes-daemonset:elasticsearch` images in my k8s cluster with default `fluent.conf` and `kubernetes.conf`. However, the json format log fields doesn't get parsed into output record fields when sending to elasticsearch.

am I missing some configurations ?

just a note, i have another older cluster running with `fluent/fluentd-kubernetes-daemonset:v0.12-alpine-elasticsearch` image that was pull couple month prior is parsing the json log fine.

thanks

repeatedly:
You need to put filter parser in your pipeline: https://docs.fluentd.org/v1.0/articles/filter_parser
See https://github.com/fluent/fluentd/issues/2073 for more details.

We will add README about this point.

lentzi90:
To help others coming here looking for solutions: Note that there are two different versions for fluentd and that the configuration syntax has changed for the filter plugin.
Compare [v0.12](https://docs.fluentd.org/v0.12/articles/filter_parser) and [v1.0](https://docs.fluentd.org/v1.0/articles/filter_parser).

[This comment](https://github.com/fluent/fluentd-kubernetes-daemonset/issues/273#issuecomment-466510424) worked for me with `v1.4.2-debian-elasticsearch-1.0`, but you may be interested in  [this](https://github.com/fluent/fluentd/issues/2073#issuecomment-409469227) if you run some other version.

ericln:
@lentzi90 thanks, that worked for me too by adding the following

```
<filter kubernetes.var.log.containers.**>
  @type parser
  format json
  key_name log
  reserve_time true
  reserve_data true
  emit_invalid_record_to_error false
</filter>
```

Issue #174, opened 2018-07-12
Make the fluentd conf configurable from configMap


vidhy:
We can have the config files retrieved from configMap ,Instead of baking the fluentd conf files inside the docker image. This would eneble user to give custom configurations.

vpistis:
In my installation I replace config files using a config map mounted into the fluent config path like so:

```
volumeMounts:
  - name: config-volume
    mountPath: /fluentd/etc
```
and for the volume:
```
volumes:
  - name: config-volume
    configMap:
      name: fluentd-config
```

tuchodi:
@vpistis 

> I replace config files

will fail in k8s, because of:
1. `configMap` is RO
https://github.com/kubernetes/kubernetes/issues/60814
https://github.com/kubernetes/kubernetes/issues/62099#issuecomment-378770047

> Fix impact:
> Secret, configMap, downwardAPI and projected volumes will be mounted as read-only volumes. Applications that attempt to write to these volumes will receive read-only filesystem errors.

2. attempt to `sed` here https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v0.12/alpine-elasticsearch/entrypoint.sh#L6-L12
```bash
if [ -z ${FLUENT_ELASTICSEARCH_USER} ] ; then
   sed -i  '/FLUENT_ELASTICSEARCH_USER/d' /fluentd/etc/${FLUENTD_CONF}
fi

if [ -z ${FLUENT_ELASTICSEARCH_PASSWORD} ] ; then
   sed -i  '/FLUENT_ELASTICSEARCH_PASSWORD/d' /fluentd/etc/${FLUENTD_CONF}
fi
```
3. will crash with error:
```
sed: can't create temp file '/fluentd/etc/fluent.confXXXXXX': Read-only file system
```

the issue is still open for k8s

vpistis:
@tuchodi I'm sorry, but I'm not specified that I used the `configMap` for fluentd 1.2.2 version, and I don't see the RO error.

tuchodi:
@vpistis 
`configMap` clearly specified here https://github.com/fluent/fluentd-kubernetes-daemonset/issues/174#issuecomment-405254408:

>and for the volume:
>
>```
>volumes:
>  - name: config-volume
>    configMap:
>      name: fluentd-config
> ```

[...]

PR #240, opened 2018-12-10
Include external configuration files

giannello:
In order to allow customizing some configuration without having to rebuild the image, we include all the configuration files in `/fluentd/etc/conf.d`

Partially solves #174
Provides a way to solve #181


repeatedly:
Could you follow DCO?

https://github.com/apps/dco

repeatedly:
Thx!

PR #242, opened 2018-12-12
Regenerate templates for fluentd

goruha:
## What
* Regenerate `fluentd.conf` from template for `v0.12-debian-elasticsearch`

## Why
* To support https://github.com/fluent/fluentd-kubernetes-daemonset/pull/202


repeatedly:
Thanks!

PR #258, opened 2019-01-30
Update to v1.3.3

repeatedly:
- Follow new image change, e.g. use tini instead of dumb-init
- Add FLUENT_ELASTICSEARCH_SED_DISABLE env var to disable sed execution
- Add FLUENTD_SYSTEMD_CONF env var to disable `in_systemd` execution with "disable"
- Add image version to tag to avoid inplace image update

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,19 +10,19 @@
 
 ### Debian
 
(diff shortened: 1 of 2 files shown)


repeatedly:
If update docker image or gems, increment major.minor version.

PR #262, opened 2019-02-04
Add daemonset to simply forward logs

keigohtr:
Hi.

Since we miss the simply log forwarding options, I made a pull-request. This option is useful for users who use a single log aggregation server which do several tasks (forward to other servers, dump logs, etc...).


repeatedly:
Your commit includes `plugins/v1.1/forward/.gitkeep` like directories.
Do these directories generate by default?
We have already dropped v1.1 / v1.2 update.

keigohtr:
Hi @repeatedly, thank you for reviewing.

> Do these directories generate by default?

No. I added them to follow the other style.
Shall I delete them?

repeatedly:
Sorry for the late. I will merge this patch after back to home.

PR #268, opened 2019-02-18
Add multi-format-parser and concat plugins

repeatedly:
These plugins are light-weight so adding it to images doesn't impact image size.

Signed-off-by: Masahiro Nakagawa <repeatedly@gmail.com>


bdashrad:
this is very helpful, thanks!

NareshAnaparthi:
where can i see Dockerfile or Image path to pull . Not sure if iam missing. Thank you

PR #276, opened 2019-02-28
Fix Encoding for Graylog output

### LINKED ISSUE #145: Sending logs to Graylog
There is no example of Graylog DaemonSet, I tried to create my own by modifying elasticsearch example. But I receive the following error:

`[out_graylog] failed to flush the buffer. error_class="Encoding::UndefinedConversionError" error="\"\\xC2\" from ASCII-8BIT to UTF-8" plugin_id="out_graylog"`

I also noticed that `protocol` is not configured in fluentd.conf, so I tested UDP and TCP with the same result.

`fluentd-daemonset-graylog.yaml` I created:
```
kind: DaemonSet
metadata:
  name: fluentd
  namespace: kube-system
  labels:
    k8s-app: fluentd-logging
    version: v1
    kubernetes.io/cluster-service: "true"
spec:
  template:
    metadata:
      labels:
        k8s-app: fluentd-logging
        version: v1
        kubernetes.io/cluster-service: "true"
    spec:
      tolerations:
      - key: node-role.kubernetes.io/master
        effect: NoSchedule
      containers:
      - name: fluentd
        image: fluent/fluentd-kubernetes-daemonset:v1.2-debian-graylog
        env:
          - name:  FLUENT_GRAYLOG_HOST
            value: "logs-graylog.default.svc.cluster.local"
          - name:  FLUENT_GRAYLOG_PORT
            value: "12201"
        resources:
          limits:
            memory: 200Mi
          requests:
            cpu: 100m
            memory: 200Mi
        volumeMounts:
        - name: varlog
          mountPath: /var/log
        - name: varlibdockercontainers
          mountPath: /var/lib/docker/containers
          readOnly: true
      terminationGracePeriodSeconds: 30
      volumes:
      - name: varlog
        hostPath:
          path: /var/log
      - name: varlibdockercontainers
        hostPath:
          path: /var/lib/docker/containers
```

Any idea?
Thanks

zrochler:
The gelf plugin used for graylog assumes `UTF-8` from the source, but the input sources are using `ASCII-8BIT`. This currently causes errors when trying to flush the buffer (See: https://docs.fluentd.org/v1.0/articles/faq#i-got-enconding-error-inside-plugin.-how-to-fix-it). To solve this I added the plugin here: https://github.com/repeatedly/fluent-plugin-record-modifier and added a filter just before the output to change the encoding.

I considered adding the encoding parameter to the sources, but that adds quite a bit more to maintain.

This fixes #145


repeatedly:
Thanks!

PR #282, opened 2019-03-22
Add FLUENT_ELASTICSEARCH_RELOAD_ON_FAILURE envvar to suppress unstable sniffering

cosmo0920:
This is reported in
https://github.com/uken/fluent-plugin-elasticsearch/issues/525.

Invalid sniffer information is obtained by default, but we can avoid
the following configuration:

```aconf
reload_connections false
reconnect_on_error true
reload_on_failure true
```

To specify reload_on_failure on fluentd-kubernetes-daemonset,
we should introduce a new envver to specify it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@
 
 IMAGE_NAME := fluent/fluentd-kubernetes
 ALL_IMAGES := \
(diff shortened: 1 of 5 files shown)


repeatedly:
Thx!

PR #285, opened 2019-03-26
Expose retention policy for CloudWatch.

jonathanjuursema:
Exposes the retention policy for CloudWatch (`retention_in_days`) via an environment variable (`RETENTION_IN_DAYS`) so it can be configured when deploying the daemonset.

### DIFF
--- a/docker-image/v1.3/debian-cloudwatch/conf/fluent.conf
+++ b/docker-image/v1.3/debian-cloudwatch/conf/fluent.conf
@@ -12,7 +12,7 @@
   log_group_name "#{ENV['LOG_GROUP_NAME']}"
   auto_create_stream true
   use_tag_as_stream true
-  retention_in_days "#{ENV['RETENTION_IN_DAYS']}"
+  retention_in_days "#{ENV['RETENTION_IN_DAYS'] || 'nil'}"
   json_handler yajl # To avoid UndefinedConversionError
   log_rejected_request "#{ENV['LOG_REJECTED_REQUEST']}" # Log rejected request for missing parts
 </match>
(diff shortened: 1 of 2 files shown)


repeatedly:
When envvar is not specified, what is the default value?

jonathanjuursema:
@repeatedly the default behavior, as specified [here](https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs) should be to set no expiry date. This is also the current behavior.

luukd:
@repeatedly any news on this PR? I'm also looking for an option to create the loggroup with a custom retention policy.

repeatedly:
`retention_in_days ` parameter is integer type. When you pass "", this happens when `RETENTION_IN_DAYS` is not set, to integer, its value is 0. This is not the problem on existing user environment? No error happen with CWL?

jonathanjuursema:
I've ran some more tests and verified that when the environment variable is not set, the cloudwatch plug-in does not put a retention policy on the log group.

In any case, existing users won't have trouble because the plug-in [only sets the retention policy when creating the log group, and doesn't update it afterwards](https://github.com/fluent-plugins-nursery/fluent-plugin-cloudwatch-logs/blob/51eccdb2a6f22951e542bcef22ebbd769bdcf663/lib/fluent/plugin/out_cloudwatch_logs.rb#L400-L410).

Regardless, I've updated the PR to more explicitly default to "no retention policy". As far as I've been able to test, this behaves the same as the original PR: 
- When omitting the environment variable, no retention policy is set.
- When specifying an environment variable, the retention policy is set only on log *group* creation.
- When an [invalid retention policy](https://docs.aws.amazon.com/AmazonCloudWatchLogs/latest/APIReference/API_PutRetentionPolicy.html#CWL-PutRetentionPolicy-request-retentionInDays) is specified, the plug-in returns an error instead of notifying the user gracefully, however, this is the behavior of the plug-in.

repeatedly:
> "#{ENV['RETENTION_IN_DAYS'] || 'nil'}"

This doesn't set `nil` because `"#{nil}"`  generates empty string and it is same as 0.
But no problem on existing user environment, we can merge this.

repeatedly:
This change will be included in v1.7 images, not v1.3.

jonathanjuursema:
Ah, thanks for the clarification and merging!

PR #287, opened 2019-04-01
Adding Prometheus metrics support

max-rocket-internet:
Fixes https://github.com/fluent/fluentd-kubernetes-daemonset/pull/97 https://github.com/fluent/fluentd-kubernetes-daemonset/issues/93
Closes https://github.com/fluent/fluentd-kubernetes-daemonset/pull/198

In this PR I am adding:
- Basic prometheus metrics support to expose fluentd internal metrics (buffers, errors, etc)
- Can be disabled by setting `FLUENTD_PROMETHEUS_CONF` env var to `disabled`

I haven't updated all the autogenerated stuff in `docker-image/v1.3/*` yet so we can see easily what I'm changing in this PR first.

Questions:
1. Should it be disabled by default?
2. Have I updated the Makefile correctly?


max-rocket-internet:
I have tested it in our environment with the cloudwatch image and it works well 😄

templates/Gemfile.erb
```diff
@@ -12,6 +12,7 @@ gem "oj", "3.5.1"
 gem "fluent-plugin-multi-format-parser", "~> 1.0.0"
 gem "fluent-plugin-concat", "~> 2.3.0"
 gem "fluent-plugin-grok-parser", "~> 2.5.0"
+gem "fluent-plugin-prometheus", "1.3.0"
```
repeatedly: `~>` is better than fixed number because teeny version is increamented for bug fixes.

max-rocket-internet: OK done!

max-rocket-internet:
@repeatedly I've addressed that comment and added all the generated template files now

repeatedly:
Thanks!
Lastly, could you update image version of Makefile like this commit?

https://github.com/fluent/fluentd-kubernetes-daemonset/commit/9b9aec0b199b93e54e3978794e995fb29104bea5#diff-b67911656ef5d18c4ae36cb6741b7965

This needs to avoid in-place image update.

max-rocket-internet:
> Lastly, could you update image version of Makefile like this commit?

OK [done](https://github.com/fluent/fluentd-kubernetes-daemonset/pull/287/commits/0bfb9ebbc6f41e6c9bb1431abfe9e1b1a414ae53). I've just bumped each version number. I hope that's correct.

repeatedly:
Thanks. just merged. Will rebuild images on docker hub.

PR #298, opened 2019-04-10
Added Datadog integration

sebastianrosch:
This PR adds Datadog integration. This has been proposed before but I couldn't get the other branches to work.

I'm still in contact with Datadog about the best way to provide all log metadata, as the format is different. For example, I am trying to parse the log level as `status` from the message and the `stream` field.


dmarkey:
Any progress on this?

PR #299, opened 2019-04-10
Allow to set protocol for Syslog Output in fluentd

stuh84:
Allow to define TCP or UDP for the syslog protocol within the Fluent image. Syslog is capable of running on TCP or UDP, so seems prudent to allow changing the protocol type.


docker-image/v1.3/debian-syslog/conf/fluent.conf
```diff
@@ -14,7 +14,7 @@
   port "#{ENV['SYSLOG_PORT']}"
   severity debug
   tag fluentd
-  protocol tcp
+  protocol "#{ENV['SYSLOG_PROTOCOL']}"
```
repeatedly: Could you add `|| 'tcp'` for default value?

stuh84: Thanks @repeatedly, this has now been updated

repeatedly:
Thx!

PR #301, opened 2019-04-17
Update fluent-plugin-elasticsearch to v3.4.2

cosmo0920:
Because previous ES plugin handles wrongly type_name warnings.

Originally, it is reported in
https://github.com/uken/fluent-plugin-elasticsearch/issues/574.


haoyujiang1994:
Thanks for your reply.

How could I update fluent-plugin-es to v3.4.2 on gcp by setting DaemonSet for fluent ? Would I just need to change fluent image from v1.3-debian-elasticsearch to v1.3.3-debian-elasticsearch ?

repeatedly:
Okay. Just merged.

haoyujiang1994:
Thank you so much for the help ! 

I also would like to know that should I write the config map myself or update the env variable in daemonSet with default config if I would like to precise the time format to nano second ?

PR #303, opened 2019-04-20
kafka is used with zookeeper

pyama86:
Hi!
I want to use `fluent-plugin-zookeeper`  because I want to write such a configuration.

```ruby
<match **>
  @type kafka_buffered
  zookeeper zookeeper.service.consul:2181
  zookeeper_path /cluster/brokers/ids
  output_include_tag true
  output_include_time true
</match>
```
Please study my suggestion.


pyama86:
> Do you want to install zookeeper gem to use zookeeper and zookeeper_path option in fluent-plugin-kafka?

yes!It is necessary to use zookeeper.

https://github.com/fluent/fluent-plugin-kafka

> If you want to use zookeeper related parameters, you also need to install zookeeper gem. zookeeper gem includes native extension, so development tools are needed, e.g. ruby-devel, gcc, make and etc.

ganmacs:
So, Why not install only zookeeper gem? I think installing zookeeper gem is enough to achieve what you want.

pyama86:
oh! sorry.
I had a misunderstanding. to correct.

pyama86:
Let me ask you one question.
Gemfile.lock updated automatically at release?

ganmacs:
> Gemfile.lock updated automatically at release?

I don't think so. `make src-all` would update all Gemfile.lock  (Is this right @repeatedly ?).

pyama86:
I update Gemfile.lock.
Is there anything else you need?

repeatedly:
No need more task. Thanks. New image will be published soon.

pyama86:
thanks!

PR #305, opened 2019-04-26
Extend environment variable for syslog image

### LINKED ISSUE #271: fluentd syslog output plugin to send logs to remote syslog server with TLS
I am new to fluentd and am able to collect logs from Kubernetes cluster, however I am hoping to send logs to a remote syslog server with ssl and cert file.

Is there a TLS supported output syslog implementation right now?

tongpu:
Primary goal is to allow configuration of TLS parameters, but
additionally all other relevant variables can now also be defined via
environment variables.

Fixes #271

### DIFF
--- a/docker-image/v1.3/debian-syslog/conf/fluent.conf
+++ b/docker-image/v1.3/debian-syslog/conf/fluent.conf
@@ -12,9 +12,12 @@
   @id out_kube_remote_syslog
   host "#{ENV['SYSLOG_HOST']}"
   port "#{ENV['SYSLOG_PORT']}"
-  severity debug
-  tag fluentd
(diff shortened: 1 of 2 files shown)


tongpu:
fluent-plugin-kubernetes_remote_syslog needs to roll a more recent release to enable the functionality, because it's not included in 0.3.6. :(

repeatedly:
Need this change?

tongpu:
@repeatedly That would be OK too. I've lost track how many forks of the remote_syslog plugin exist currently, but I would need one that allows me to specify the `tls` and `ca_file` configuration at least. The changes in #211 would help with getting this PR merged.

repeatedly:
Close via #306

PR #306, opened 2019-04-30
use fluent-plugin-remote_syslog for fluentd v1.x

tongpu:
Additionally this enables the use of TCP+TLS for syslog transport.

This makes #305 obsolete and implements the same change for v1.3 as #211


tongpu:
@repeatedly What do you think about that as a solution for both #305 and #211 ?

repeatedly:
Good. Thanks!

PR #311, opened 2019-05-08
Use Secret for Logentries

tokyowizard:
Since the data contains a token, `Secret` should be used instead of `ConfigMap`.

Signed-off-by: Frank Lee <frank@cryptact.com>


tokyowizard:
I've had to move on from Logentries, so will be closing this PR.

PR #315, opened 2019-05-13
Update fluent-plugin-elasticsearch to v3.5.1

Nuru:
This PR updates Elasticsearch support

- Update [fluent-plugin-elasticsearch](https://github.com/uken/fluent-plugin-elasticsearch) to pick up https://github.com/uken/fluent-plugin-elasticsearch/pull/586 which implements a configurable limit on message size.
- Pin `elasticsearch` Ruby gem to version 6.x. Version 7.0.0 is out and is not expected to be compatible with ES 6.x, but 6.x I still what most people are using.
- Add `FLUENT_ELASTICSEARCH_BULK_MESSAGE_REQUEST_THRESHOLD` environment variable to configure the maximum allowed size of a request
- Add `FLUENT_ELASTICSEARCH_RETRY_FOREVER` to allow people to turn off `retry_forever`


templates/conf/fluent.conf.erb
```diff
@@ -32,13 +32,14 @@
    index_name "#{ENV['FLUENT_ELASTICSEARCH_LOGSTASH_INDEX_NAME'] || 'logstash'}"
    type_name "#{ENV['FLUENT_ELASTICSEARCH_LOGSTASH_TYPE_NAME'] || 'fluentd'}"
 <% if is_v1 %>
+   bulk_message_request_threshold "#{ENV['FLUENT_ELASTICSEARCH_BULK_MESSAGE_REQUEST_THRESHOLD'] || '10M'}"
```
ganmacs: [nits] Default value looks 20M (https://github.com/uken/fluent-plugin-elasticsearch/blob/v3.5.1/lib/fluent/plugin/out_elasticsearch.rb#L56). Was this done on purpose?

Nuru: Yes, done on purpose, though now that I know more about it I would default it to 4M. 20M seems to have been set somewhat arbitrarily, whereas I am concerned about the 10M request size limit on AWS Elasticsearch. See https://github.com/uken/fluent-plugin-elasticsearch/issues/588 for details.

docker-image/v1.4/debian-elasticsearch/Gemfile
```diff
@@ -10,7 +10,8 @@ gem "fluent-plugin-concat", "~> 2.3.0"
 gem "fluent-plugin-grok-parser", "~> 2.5.0"
 gem "fluent-plugin-prometheus", "~> 1.3.0"
 gem "fluent-plugin-rewrite-tag-filter", "~> 2.1.0"
-gem "fluent-plugin-elasticsearch", "~> 3.4.2"
+gem "elasticsearch", "~> 6.0"
```
ganmacs: I'd have to say that it's a breaking change to change the elasticsearch version from 7.0.0 to 6.x.x even if many people are using an older version. It looks better for me to change the image name to `v1.4.2-debian-elasticsearch-2.0`.

Do you have any idea? @repeatedly

Nuru: I view it the other way around. AFAIK v1.4.2-debian-elasticsearch-1.0 was never actually released, or if it was released, was only released in the past few hours. Prior to that, [only elasticsearch 6.x plugins](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/8c76f51696bdeea7643ec0e0696afebf336c45a9/docker-image/v1.3/debian-elasticsearch/Gemfile.lock#L16-L21) were used. To me, it was the unspecified use of "latest" elasticsearch pulling in the 7.0.0 plugin that was the unexpected breaking change.

repeatedly: One idea is separating tags into xx-debian-elasticsearch7-1 and xx-debian-elasticsearch6-1.

Nuru: While we are at it, we should also consider that people are asking for a version that uses https://github.com/atomita/fluent-plugin-aws-elasticsearch-service

However, for now, the easiest first step is to just accept this PR as written and then add separate support for ES 7 and the AWS plugin later.

repeatedly:
https://github.com/fluent/fluentd-kubernetes-daemonset/tree/master/docker-image/v1.7
Since v1.7, elasitcsearch images are separated into 6 and 7.

PR #316, opened 2019-05-23
Add Loggly example

heidemn:


heidemn:
The RBAC example works for me (EKS, Kubernetes 1.12),
For non-RBAC example, at least "kubectl apply" succeeded. It should work in theory on a cluster without RBAC.

heidemn:
@repeatedly are there any changes needed before this can be merged?
We're running exactly this RBAC example in two production clusters, with no changes except for the Loggly token.

repeatedly:
Thanks for the examples. Just merged.

Issue #293, opened 2019-04-04
How to maintain contirubted images


repeatedly:
We have contributed images and pending PR for new destination.
Currently, we have a problem for the maintenance.

1. The repository authors don't know the knowledge of some destinations, e.g. logzio, stackdriver, kinesis, so we can't reply the answer to some issue and questions. In addition, we can't judge the patch is correct or not for these destinations.
2. Merging patch and releasing new images for contributed images are delayed because we need to wait the patch from contributors. This is not good for users.

We want to know how to avoid these problems.  One idea is adding contributors to this repository but it increases security concern like backdoored images.
Any feedback is welcome.

perk:
Hi @repeatedly it seems like there are three categories of images:

1. opensource projects, like elasticsearch, kafka or just fluentd (forwarder)
2. destinations from big players and huge userbases (like stackdriver or cloudwatch)
3. other destinations, like papertrail, logzio or still waiting logsense :wink:

Dropping 2 and 3 will make things easier to maintain for sure but will also create a barrier for newcomers.

The 2nd category is tricky because people are interested in using them but unfortunately companies may not be interested in maintaining all the projects out there (this repo included).
I'd keep them for the sake of making things easier for new users.

The least tricky is in my opinion the 3rd category - maybe there should be a policy that other destinations need to have at least two maintainers?
If then nobody is interested in resolving issues for 3rd party proprietary destinations then I think it is fair to just drop them.

I do understand why you think about dropping 2 and 3 altogether. In such case we (@logsenseapp) will just maintain our own forked repo.

perk:
Hi @repeatedly it seems like there are three categories of images:

1. opensource projects, like elasticsearch, kafka or just fluentd (forwarder)
2. destinations from big players and huge userbases (like stackdriver or cloudwatch)
3. other destinations, like papertrail, logzio or (still waiting) logsense :wink:

I'd keep the first category.

Dropping 2 and 3 will make things easier to maintain for sure but will also create a barrier for newcomers.

The 2nd category is tricky because people are interested in using them but unfortunately companies may not be interested in maintaining all the projects out there (this repo included).
I'd keep them for the sake of making things easier for new users.

The least tricky is in my opinion the 3rd category - maybe there should be a policy that other destinations need to have at least two maintainers?
If then nobody is interested in resolving issues for 3rd party proprietary destinations then I think it is fair to just drop them.

[...]

PR #317, opened 2019-05-27
Add AWS Elasticsearch image

AdrienKuhn:
This PR adds `debian-awselasticsearch` docker image. 

AWS ES service require requests signature. 

This image contains the [aws-elasticsearch-service fluent](https://github.com/atomita/fluent-plugin-aws-elasticsearch-service) plugin.

Authentication with AWS ES is done using EC2 instance profile.


Jaykah:
Is anyone using this in production?

Also, any plans to merge?

Thanks!

repeatedly:
Currently, no plan to merge because we have a problem to maintain contributed images.

https://github.com/fluent/fluentd-kubernetes-daemonset/issues/293

danopia:
I've been using an older version of the same thing in production for at least a year now. Just switched production to this image because [I needed a newer elasticsearch plugin](https://github.com/uken/fluent-plugin-elasticsearch/issues/589#issuecomment-495520040). I also replace `/fluentd/etc` with my own configuration at the Kubernetes level, so that I can roll out new configs without making new images. Works well!

gatolim:
@AdrienKuhn  thanks for coming up with the repo. I am new to the docker world and was playing around this. I have my ES set up in AWS and have docker for window set up locally. I cloned your repo and docker build the debian-awselasticsearch image however when I try to apply the image using fluentd-daemonset-awselasticsearch-rbac.yml I have the following error: 
[FATAL tini (6)] exec /fluentd/entrypoint.sh failed: No such file or directory

I verified that the build steps does copy this .sh file and the file existed in my build folder. Am I missing something ?

AdrienKuhn:
@gatolim I tried to apply the `fluentd-daemonset-awselasticsearch-rbac.yml` file on my side and I didn't encounter any issue. 

I used the following (v1.7) image for test purposes: [krewh/fluentd-awselasticsearch](https://cloud.docker.com/repository/docker/krewh/fluentd-awselasticsearch/general).

gatolim:
Thanks @AdrienKuhn, I ended up building my own image. I encountered the same problem while using my own and it turns out to be the end line char ('\r\n') difference between window and linux in the .sh file. I suspected it was the same cause when I use yours.

wcarlsen:
We ended up using this proxy (https://github.com/abutaha/aws-es-proxy) as a sidecar, which allows us to use on of the pre built images of this repos. Works flawlessly, but the proxy image development is rather stale.

Nuru:
> We ended up using this proxy (https://github.com/abutaha/aws-es-proxy) as a sidecar, which allows us to use on of the pre built images of this repos. Works flawlessly, but the proxy image development is rather stale.

@wcarlsen Did you really get `aws-es-proxy` to work flawlessly? Last time I tried it I had to reject it because if it encountered a 403 or 302 status code from Elasticsearch it would just swallow it and return an empty 200 to the caller. How did you fix this, or did you just decide it was acceptable behavior?

wcarlsen:
@Nuru we probably shouldn't continue this here. But I wasn't aware of this behavior. I think it is a trade off between maintaining a fluentd image. We are currently investigating when we would expect to get a 302 or 403 from AWS ES. Maybe you have experience with this or have a better overview of how big a problem this actually is?

PR #320, opened 2019-06-10
Update fluentd-daemonset-cloudwatch-rbac.yaml to use latest image

motemen:
The image `fluent/fluentd-kubernetes-daemonset:cloudwatch` referred in the example seems to be old and not maintained, so changed it to latest one.


repeatedly:
Thanks!

PR #323, opened 2019-06-12
Add log_es_400_reason parameter to ES config

kevinscholz:
To be able to use the fluentd-elasticsearch image if you want to enable the log_es_400_reason option, you currently need to define the ES match block again in your own config. 

This change allows setting it via environment variable.


kevinscholz:
I just noticed, that this option is not available in v1.18.0 which is used by the v0.12 image. Checking if fluentd just ignores it or if it would break the v0.12 image.

kevinscholz:
Since only v1.4 of the image is updated via the Makefile, we should be fine. Looking forward to your feedback.

repeatedly:
Thanks!

Issue #283, opened 2019-03-23
Enable protocol for graylog


newbenji:
Make sure protocol can be set like host and port can be set.
     With a default for udp, if not set

 - name: FLUENT_GRAYLOG_PROTOCOL
         value: "tcp|udp"

cosmo0920:
Already merged. Closing.

PR #284, opened 2019-03-23
Make it possible to set protocol on gelf

newbenji:

PR #343, opened 2019-08-13
Made the protocol available to the graylog config

Spazzy757:
# Changelog
- Made the ENV "FLUENT_GRAYLOG_PROTOCOL" which fixes issue https://github.com/fluent/fluentd-kubernetes-daemonset/issues/283

This is useful specifically because with Graylog you need to specify using TCP/UDP so using the default UDP protocol is very limiting

### DIFF
--- a/templates/conf/fluent.conf.erb
+++ b/templates/conf/fluent.conf.erb
@@ -167,6 +167,7 @@
   include_tag_key true
   host "#{ENV['FLUENT_GRAYLOG_HOST']}"
   port "#{ENV['FLUENT_GRAYLOG_PORT']}"
+  protocol "#{ENV['FLUENT_GRAYLOG_PROTOCOL'] || 'udp'}"
   <buffer>
     flush_thread_count 8
     flush_interval 5s


Spazzy757:
I realise there is the PR that does the same but has an issue with the default https://github.com/fluent/fluentd-kubernetes-daemonset/pull/284, is there a need to have a default set?

shinebayar-g:
I want to ask that question from you. 
- Does the plugin supports protocol choice? 
- Does it have defaults? 

Also you're direct editing auto generated config file. Not template file.

Spazzy757:
> I want to ask that question from you.
> 
> * Does the plugin supports protocol choice?
> * Does it have defaults?
> 
> Also you're direct editing auto-generated config file. Not a template file.

The plugin does support specifying a protocol (namely it supports `TCP` or `UDP`), reason being is we had the need to specify TCP in our system, I have removed my changes from the auto-generated file and put them into the `fluent.conf.erb`.  Sorry I also specified the UDP default

repeatedly:
Don't change indent size for feature request patch.

Spazzy757:
> Don't change indent size for feature request patch.

Fixed the indentation

repeatedly:
Thanks. New image will be available soon.

PR #350, opened 2019-09-16
add missing fluentd elasticsearch user password remove hack

dbeal-wiser:
Not sure if this was intentional but seems that the user/password line remove hack was lost with the 1.7 fluentd release.

Signed-off-by: Daniel Beal <dbeal@wiser.com>


repeatedly:
Good catch... but this patch is incomplete.
You need to update original file, not generated file.
For example, `target.include?("elasticsearch")`.

https://github.com/fluent/fluentd-kubernetes-daemonset/blob/d33872f9b5fb047e795c5c2c9ad3a63efc535788/templates/entrypoint.sh.erb#L4

dbeal-wiser:
ok generated file was updated.

repeatedly:
Thanks!

PR #357, opened 2019-10-07
Changed TLS version

Oscar3170:
I could be wrong but I think elasticsearch7.0 and up only accepts TLS1.1 or more because i just had issues with that, so is there any reason to leave the default as TLS1 here?


repeatedly:
Could you follow DCO? Check "DCO details".

PR #349, opened 2019-09-16
adding fluent-plugin-detect-exceptions plugin to fluentd-elasticsearch image…

shivpathak:
Signed-off-by: Shiv Pathak <shivpathak23@gmail.com>

PR #358, opened 2019-10-10
add fluent-plugin-detect-exceptions to all images

essh:
PR #349 added fluent-plugin-detect-exceptions to the elasticsearch images. This PR adds it across all images as it is generally quite useful.


repeatedly:
Sorry for the delay. Looks good.

PR #371, opened 2019-11-17
added required service role and rbac info into yaml

scanlonjts:
merged creating service  account , rbac  and binding into yaml.


repeatedly:
Thanks!

Issue #363, opened 2019-10-26
error: tlsv1 alert protocol version (OpenSSL::SSL::SSLError)


masterkain:
can we make

```
          - name:  FLUENT_ELASTICSEARCH_SSL_VERIFY
            value: "false"
          - name: FLUENT_ELASTICSEARCH_SSL_VERSION
            value: "TLSv1_2"
```

env values more prominent in the https://docs.fluentd.org/container-deployment/kubernetes guide?

![Screenshot 2019-10-26 at 10 14 35](https://user-images.githubusercontent.com/12844/67616468-7be3c780-f7d9-11e9-9451-ac8d8eccd237.png)

took me a good while to find for connecting successfully to a quickstart elasticsearch: https://www.elastic.co/guide/en/cloud-on-k8s/1.0/k8s-quickstart.html

cosmo0920:
@ganmacs Can we close this?
https://github.com/fluent/fluentd-kubernetes-daemonset/pull/376 and https://github.com/fluent/fluentd-docs-gitbook/pull/118 are already merged.

ganmacs:
Right! thank you!

PR #376, opened 2019-12-02
Add options for TLS version and verify explicitly

cosmo0920:
Because recent Elasticsearch clusters usually use TLSv1.2,
however, fluent-plugin-elasticsearch's default value of SSL/TLS version
is TLSv1.

ref: https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/docker-image/v1.7/debian-elasticsearch7/conf/fluent.conf#L20

Related to https://github.com/fluent/fluentd-kubernetes-daemonset/issues/363.


repeatedly:
Thanks!

PR #379, opened 2019-12-10
update elasticsearch plugin for elasticsearch 7

golgoth31:
This PR will update the elasticsearch plugin to the 3.7.1 version and add the elasticsearch-xpack gem to be able to use ILM from within the plugin.


repeatedly:
Thanks!

PR #382, opened 2019-12-13
Update fluentd-daemonset-syslog.yaml

fitz0017:
added required selector field and updated apiversion to apps/v1


repeatedly:
Thanks!

PR #395, opened 2020-01-15
Adding SASL_SSL dependencies for the rdkafka plugin

naumanrana:
For SASL_SSL with GSSAPI to work with the rdkafka plugin, needed to install a bunch of kerberos related libraries and upgrade the fluent-plugin-kafka to the latest version as rdkafka was not available in 0.7.9.

### DIFF
--- a/templates/Dockerfile.erb
+++ b/templates/Dockerfile.erb
@@ -18,12 +18,13 @@ ENV FLUENTD_DISABLE_BUNDLER_INJECTION 1
 
 COPY Gemfile* /fluentd/
   RUN buildDeps="sudo make gcc g++ libc-dev libffi-dev<% if target == "graylog" %> build-essential patch zlib1g-dev liblzma-dev<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config<% end %><% if requires_git %> git<% end %>" \
-     && apt-get update \
+  runtimeDeps="<% if target == "kafka" %>krb5-kdc libsasl2-modules-gssapi-mit libsasl2-dev <% end %>" \
+     <% if target == "kafka" %> && export DEBIAN_FRONTEND=noninteractive <% end %> && apt-get update \
      && apt-get upgrade -y \
      && apt-get install \
      -y --no-install-recommends \
-     $buildDeps net-tools \
-    && gem install bundler --version 1.16.2 \
+     $buildDeps $runtimeDeps net-tools \
+    && gem install bundler --version 2.1.2 \
     && bundle config silence_root_warning true \
     && bundle install --gemfile=/fluentd/Gemfile --path=/fluentd/vendor/bundle \
     && SUDO_FORCE_REMOVE=yes \
@@ -32,7 +33,9 @@ COPY Gemfile* /fluentd/
                   $buildDeps \
  && rm -rf /var/lib/apt/lists/* \
     && gem sources --clear-all \
-    && rm -rf /tmp/* /var/tmp/* /usr/lib/ruby/gems/*/cache/*.gem
+    && rm -rf /tmp/* /var/tmp/* /usr/lib/ruby/gems/*/cache/*.gem <% if target == "kafka" %> \
+    && ldd $(gem contents rdkafka | grep librdkafka.so) | grep libsasl2.so.2 
+    <% end %> 
 
 # Copy configuration files
 COPY ./conf/fluent.conf /fluentd/etc/
--- a/templates/Gemfile.erb
+++ b/templates/Gemfile.erb
@@ -91,7 +91,8 @@ gem "fluent-plugin-remote_syslog"
 gem "fluent-plugin-kubernetes_remote_syslog"
 <% end %>
 <% when "kafka" %>
-gem "fluent-plugin-kafka", "~> 0.7.9"
+gem "rdkafka", "~> 0.7.0"
+gem "fluent-plugin-kafka", "~> 0.12.2"
 gem "snappy", "~> 0.0.15"
 <% when "kinesis" %>
 <% if is_v1 %>


templates/Dockerfile.erb
```diff
@@ -17,22 +17,30 @@ ENV GEM_HOME /fluentd/vendor/bundle/ruby/2.6.0
...
-    apt-get purge -y --auto-remove \
-                  -o APT::AutoRemove::RecommendsImportant=false \
-                  $buildDeps \
- && rm -rf /var/lib/apt/lists/* \
-    && gem sources --clear-all \
-    && rm -rf /tmp/* /var/tmp/* /usr/lib/ruby/gems/*/cache/*.gem
+
+RUN buildDeps="sudo make gcc g++ libc-dev libffi-dev<% if target == "graylog" %> build-essential patch zlib1g-dev liblzma-dev \
+<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config python <% end %><% if requires_git %> git<% end %>" \
+runtimeDeps="<% if target == "kafka" %>krb5-kdc libsasl2-modules-gssapi-mit libsasl2-dev <% end %> " \
```
repeatedly: Please move to `<% elsif target == "kafka" %>`

naumanrana: I understand it seems redundant but I did it this way in case there are additional buildDeps for other targets in the future or if other targets have some runtimeDeps in the future. Thought it would be easier to modify this way. 

We are also installing the libraries under $buildDeps $runtimeDeps net-tools (lines 27-29). Otherwise, will need an if block there to include the runtimeDeps only for kafka (even though it'll still work if the $runDeps variable does not exist technically).

Thoughts? Should I still continue to move it to <% elsif target == "kafka" %>?

elafontaine: Hi @repeatedly ,

We're trying to contribute the mechanics for adding runtime dependencies like libsasl that needs to be present with librdkafka for getting the Kerberos Authentification working with Kafka.

As the builddeps are removed as expected at the end, any way to keep the librairies present after?

Regards

elafontaine: :)

elafontaine: Thanks, feedback is appreciated as always :)

templates/Dockerfile.erb
```diff
@@ -17,22 +17,30 @@ ENV GEM_HOME /fluentd/vendor/bundle/ruby/2.6.0
...
+   && bundle install --gemfile=/fluentd/Gemfile --path=/fluentd/vendor/bundle \
+   && SUDO_FORCE_REMOVE=yes \
+   apt-get purge -y --auto-remove \
+         -o APT::AutoRemove::RecommendsImportant=false \
+         $buildDeps \
+   && rm -rf /var/lib/apt/lists/* \
+   && gem sources --clear-all \
+   && rm -rf /tmp/* /var/tmp/* /usr/lib/ruby/gems/*/cache/*.gem \
+   <% if target == "kafka" %> \
+   && ldconfig -p | grep -q "libsasl2.so$" \
```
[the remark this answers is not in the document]
repeatedly: @naumanrana How about this point?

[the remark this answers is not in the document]
naumanrana: @elafontaine is correct. We have modified it to use ldd instead to see if its linked against the sasl dependency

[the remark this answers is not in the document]
elafontaine: Nauman, you can close this comment (I can't on my side)

templates/Dockerfile.erb
```diff
@@ -17,22 +17,28 @@ ENV GEM_HOME /fluentd/vendor/bundle/ruby/2.6.0
...
+<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config python <% end %><% if requires_git %> git<% end %>" \
+runtimeDeps="<% if target == "kafka" %>krb5-kdc libsasl2-modules-gssapi-mit libsasl2-dev <% end %> " \
+   && export DEBIAN_FRONTEND=noninteractive \
+   && apt-get update \
+   && apt-get upgrade -y \
+   && apt-get install \
+   -y --no-install-recommends \
+   $buildDeps $runtimeDeps net-tools \
+   && gem install bundler --version 2.1.2 \
+   && bundle config silence_root_warning true \
```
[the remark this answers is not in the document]
elafontaine: I realise that this didn't print until now...

templates/Dockerfile.erb
```diff
@@ -17,13 +17,15 @@ ENV GEM_HOME /fluentd/vendor/bundle/ruby/2.6.0
 ENV FLUENTD_DISABLE_BUNDLER_INJECTION 1
 
 COPY Gemfile* /fluentd/
-  RUN buildDeps="sudo make gcc g++ libc-dev libffi-dev<% if target == "graylog" %> build-essential patch zlib1g-dev liblzma-dev<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config<% end %><% if requires_git %> git<% end %>" \
+  RUN buildDeps="sudo make gcc g++ libc-dev libffi-dev<% if target == "graylog" %> build-essential patch zlib1g-dev liblzma-dev<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config python <% end %><% if requires_git %> git<% end %>" \
```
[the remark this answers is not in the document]
naumanrana: Thought it was required but its not. Removed.

elafontaine:
I think this PR is ready to be merge;
1- good commit message explaning the manipulation
2- Adds a feature for runtime dependencies
3- Adds a test mechanism for the container setup.
@repeatedly Anything left?

@naumanrana  👍 Nice Job

elafontaine:
Looks good to me

templates/Dockerfile.erb
```diff
@@ -18,12 +18,14 @@ ENV FLUENTD_DISABLE_BUNDLER_INJECTION 1
 
 COPY Gemfile* /fluentd/
   RUN buildDeps="sudo make gcc g++ libc-dev libffi-dev<% if target == "graylog" %> build-essential patch zlib1g-dev liblzma-dev<% elsif target == "kafka" %> build-essential autoconf automake libtool pkg-config<% end %><% if requires_git %> git<% end %>" \
+  runtimeDeps="<% if target == "kafka" %>krb5-kdc libsasl2-modules-gssapi-mit libsasl2-dev <% end %>" \
+     && export DEBIAN_FRONTEND=noninteractive \
```
repeatedly: Why add this envvar?
Executing `ldd $(gem contents rdkafka | grep librdkafka.so) | grep libsasl2.so.2` will be blocked?

elafontaine: No, it's krb5 which is blocking.  I had that discussion offline.

elafontaine: The commit message should be explainjng the problem.

repeatedly: Ah, I missed the commmit message. If this is only for kafka image, it should be in `<% if target == "kafka" %>`.

naumanrana: Done. Note that I have added it into a new if block as I want a complete separation from the runtimeDeps section

repeatedly:
Thanks. New images will be published soon;.

elafontaine:
Thanks :) eager to use it :).

BTW, I was looking at the issue https://github.com/fluent/fluentd-kubernetes-daemonset/issues/293
and I was wondering if what we did in this commit could be some (small part) of a bigger solution.  

I would like to get your thoughts and know if we could have done something better.

PR #398, opened 2020-01-30
Add option to set `template_overwrite` in config

eyalzek:
This defaults to false otherwise and might be inconvenient when specifying an index template.


eyalzek:
@ganmacs can we merge this? :)

ganmacs:
thanks!

repeatedly:
Just published new image with same tag: https://github.com/fluent/fluentd-kubernetes-daemonset/commit/959fce206d2b628ba6d17863ea4d2ce6f9930a05

PR #403, opened 2020-02-12
Add Elasticsearch sniffer class name and use it by default

cosmo0920:
Because elasticasearch plugin should use elasticsearch-ruby client's sniffering cluster feature which is called every 10000 requests by default and its sniffering feature replaces elasticsearch instances service names with their pod ip addresses.
But, in k8s environment, we should use service name instead of raw ip addresses.

So, in k8s, we should recommend to use SimpleSniffer class for elasticsearch plugin.
WDYT?


cosmo0920:
Let's :shipit:!

cosmo0920:
I triggers es6 and es7 image builds.
They will be published with the new tags.

PR #406, opened 2020-02-17
add include timestamp

wrender:
It seems like this setting is needed to include @timestamp information, when using settings logstash_format false, and index_name to define an index name.


repeatedly:
Thanks!

PR #419, opened 2020-03-06
Allow setting skip_* settings for k8s metadata filter using ENV vars

ojundt:
I like that the k8s metadata filter plugin is included by default however I'd also like to be able to tweak the settings without having to overwrite the complete kubernetes.conf. 
Especially the `skip_labels`, `skip_container_metadata`, `skip_master_url` and `skip_namespace_metadata` settings are of interest.
This PR enables users to set these settings with ENV vars.


repeatedly:
Sorry for the delay. Just merged.

PR #424, opened 2020-04-02
Update fluentd-daemonset-syslog.yaml

mazen-hasoun:


ganmacs:
No background, we can't understand why you need this.

repeatedly:
No response. closed.

PR #428, opened 2020-04-14
Update fluent-plugin-google-cloud to 0.8.4

nesl247:
This is significantly outdated, not sure as to the reason.

This does need to be tested.


repeatedly:
Does anyone test this version?

nesl247:
I actually tested this out and it fails. They require an older version of fluentd, so the gems don't install.

PR #431, opened 2020-04-23
Minor fixes

elsesiy:
Remove unused variables
Set bundler to 2.1.4 to match generated images
Make bash/sh scripts more portable by using `/usr/bin/env shell` instead of hard-coded `/bin/shell`


elsesiy:
Closed since I included the changes in #435

PR #433, opened 2020-04-25
Update Gemfile

ashishaffinion:
add fluent-plugin-color-stripper plugin in gems so that we can use plugin color-striper


cosmo0920:
v0.12 series should be outdated. Closing.

PR #446, opened 2020-06-03
fix: Update elasticsearch plugin

golgoth31:
Update fluent-plugin-elasticsearch to 4.09
Update elasticsearch-xpack to 7.0 (to be in sync with "elasticsearch")


repeatedly:
Thanks. It will be included in v1.11.0 images.

PR #467, opened 2020-07-20
debian-elasticsaearch7: Use fluent-plugin-elasticsearch 4.4.1

cosmo0920:

fluent-plugin-elasticsearch v4.1.1 can handle `logstash_format` as `true` and `enable_ilm` circumstances.


cosmo0920:
But I'm not sure debian-elasticsearch7 image should support ILM feature parameters....

PR #470, opened 2020-07-23
update s3 related gems

xcrezd:
We encountered on some bugs with exporting logs to s3, after manually version updating, bugs are disappeared.

`[warn]: section <web_identity_credentials> is not used in <match {kubernetes.var.log.containers.**default**.log,kubernetes.var.log.containers.**nginx-ingress**.log,kubernetes.var.log.containers.**dashboard**.log}>`


cosmo0920:
I'd published `fluent/fluentd-kubernetes-daemonset:v1.11.1-debian-s3-1.1` image. Thanks for the feedback! :muscle:

PR #487, opened 2020-08-13
Upgrade fluent-plugin-azure-storage-append-blob version

nilsga:
Upgrade fluent-plugin-azure-storage-append-blob version to 0.2.0, which
supports MSI.


elsesiy:
@nilsga I just merged the changes to the plugin. Can you please also update the `fluentd-daemonset-azureblob.yaml` example to include all available envs and the `fluent.conf.erb` accordingly? If not I can do it this weekend as this was on my list anyway.

nilsga:
Ok, I'll do that. I didn't add anything, on purpose, since the defaults were fine, and more complex configuration required custom ConfigMap configuration anyway.

elsesiy:
looks great, thanks @nilsga. Do you mind adding the `AZUREBLOB_SAS_TOKEN` as an option to the manifest as well? Then we have all options covered

cosmo0920 (approved):

cosmo0920:
I'll publish a new image tag which includes this change. Thank you for your work!

PR #489, opened 2020-08-26
Add parameter target_index_key

emrocha:
About the parameter target_index_key
Tell this plugin to find the index name to write to in the record under this key in preference to other mechanisms.
If it is present in the record (and the value is non falsy) the value will be used as the index name to write to and then removed from the record before output; if it is not found then it will use logstash_format or index_name settings as configured.
Source: https://github.com/uken/fluent-plugin-elasticsearch#target_index_key


repeatedly:
@cosmo0920 How about this change?

cosmo0920:
I think that, ideally, this is needless parameter but default parameter does not handle buffer section with user-defined attributes. I'm not against to merge this patch.

cosmo0920 (approved):

Issue #457, opened 2020-06-29
Add support for ARM64 images


Arvoreen:
I was going to try to use this on my Raspberry PI k3s kubernetes cluster, but all of the images are for x86 architecture only.  It doesn't appear to be a limitation of fluentd itself, as I see there are arm64 versions available.

Arvoreen:
FYI -- it looks like it might be as easy as switching the docker build command line in the Makefile to 
docker buildx build --platform linux/arm/v7,linux/arm64/v8,linux/amd64 <rest of arguments>

Arvoreen:
so I have been playing more, and the simple solution isn't going to work because the upstream fluentd images are not multi-architecture images.  Instead, there is a separate tag per architecture, so you can't take advantage of docker buildx multi-platform builds directly

Arvoreen:
As a finally follow up. I was able to manually build an arm64 image using an very slightly modified Dockerfile (putting the fluentd arm64 tag) that I published and then was able to successfully launch on my Raspberry Pi K3s cluster

cosmo0920:
We are building daemonset images on DockerHub. It supports non x86_64 image building, @Arvoreen ?

Arvoreen:
I could not get them to work -- I had to build a custom arm64 image.  Looking at the tag on Docker HUB shows that you are building only AMD64 images
https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/tags

scottmalkie:
It looks like 1.9 is here: https://hub.docker.com/r/arm64v8/fluentd

Could you provide multi-arch images in the main fluentd repository instead? That would make it transparent for end-users.

See https://www.docker.com/blog/multi-platform-docker-builds/ and https://tech.smartling.com/building-multi-architecture-docker-images-on-arm-64-c3e6f8d78e1c for more information.

cosmo0920:
Docker Hub does not support buildx just support amd64 image building.
We don't want to build non amd64 images in local build environment.
Currently, no way to handle to build non amd64 images.

cosmo0920:
I'd created arm64 image building triggers and kicked them.
You can use arm64 image with the listed tags:
https://github.com/fluent/fluentd-kubernetes-daemonset#arm64-images

Anyway, thank you for request to support arm64 images.

PR #491, opened 2020-09-02
Add arm64 image definitions

cosmo0920:
Fixes https://github.com/fluent/fluentd-kubernetes-daemonset/issues/457

Build under qemu-aarch64-static's aarch64 emulation. 
Can we add arm64 images build triggers on DockerHub?


repeatedly:
LGTM. We can try build arm64 images on dockerhub and get feedback from users.

repeatedly (approved):

repeatedly:
Good. For additional work, can we remove arm64 from the tag like https://hub.docker.com/r/fluent/fluent-bit/tags ?

cosmo0920:
Yep. We can do.
I guess that unifying arm64 and x86_64 image tags and build each architecture and then push them can remove arm64 from image tags.

PR #492, opened 2020-09-03
Unify arm64 and x86 64 image tags

cosmo0920:
Related to https://github.com/fluent/fluentd-kubernetes-daemonset/pull/491#issuecomment-686159433.


cosmo0920:
Umm..., DockerHub build should overwrite tag when using same tag on different architecture....

cosmo0920:
Pushing same tag with different architectures may be needed to build with docker buildx which is not supported on DockerHub....

PR #499, opened 2020-10-14
Add missing id attributes

danielfm:
This PR sets the [`@id` attribute](https://docs.fluentd.org/configuration/plugin-common-parameters#id) for all sources, filters and outputs that currently do not specify it.

This change should prevent high cardinality data in Prometheus by avoiding `object:<random-id>` values in the `plugin_id` label, as well as make the fluentd logs themselves more readable.


repeatedly:
Looks good. Thanks!

PR #514, opened 2020-11-13
Custom timekey and chunk_limit_size on S3 target

### LINKED ISSUE #513: Custom timekey and chunk_limit_size on S3 target
I noticed that it is possible to configure flush period and max chunk size for many targets, instead on S3 `timekey` and `chunk_limit_size` are hardcoded.
We need to configure these parameters (few minutes and some megabytes), can you define new environment variables to customize these values?

fvasco:
Fix #513

### DIFF
--- a/templates/conf/fluent.conf.erb
+++ b/templates/conf/fluent.conf.erb
@@ -102,9 +102,9 @@
   <buffer>
     @type file
     path /var/log/fluentd-buffers/s3.buffer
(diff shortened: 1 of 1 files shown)


repeatedly:
Could you follow DCO? Need your sign in commits.
See `Details` link above.

cosmo0920 (approved):

PR #518, opened 2020-12-11
Migrate azblob fluent plugin

elsesiy:
Replace fluent-plugin-azure-storage-append-blob with actively maintained fluent-plugin-azure-storage-append-blob-lts
Add documentation for using connection string auth


elsesiy:
@cosmo0920 @repeatedly Mind having a look at this? Thanks!

cosmo0920 (approved):

PR #521, opened 2020-12-25
Support CRI parser

repeatedly:
Based on https://github.com/fluent/fluent-plugin-parser-cri
Create `tail_container_parse.conf` to support https://github.com/fluent/fluent-plugin-parser-cri#parse-message-part-with-parsers via ConfigMap


cosmo0920 (approved):
I think that this should be included in v1.12.0-xxx-1.1 images before bumping Fluentd version to v1.12.1 or later.

andrecp:
Much appreciated!

PR #526, opened 2021-01-24
Bump fluentd to 1.12.0

elsesiy:
Add x86 and arm64 images for fluentd 1.12.0
Update fluent-plugin-kubernetes_metadata_filter to 2.6.0
Update fluent-plugin-azure-storage-append-blob-lts to 0.4.0
Update oj to 3.11.0
Update fluent-plugin-grok-parser to 2.6.2
Update fluent-plugin-prometheus to 1.8.5
Update fluent-plugin-record-modifier to 2.1.0
Update fluent-plugin-detect-exceptions to 0.0.13
Update fluent-plugin-rewrite-tag-filter to 2.4.0


elsesiy:
@repeatedly @cosmo0920 Please take a look & let me know if it needs any changes. Thanks!

cosmo0920 (approved):
LGTM.
I'm investing how to support Linux capability on these images but I didn't find it.
And the latest ruby:alpine image contains bundler 2.2 series. 
So this version of bundler starts to use `x86_64-linux-musl` platform instead of `ruby`.

Thus, we should fixed image version which is `ruby:alpine3.10` for generating Gemfile.lock, I guess.
WDYT, @repeatedly?

elsesiy:
> LGTM.
> I'm investing how to support Linux capability on these images but I didn't find it.
> And the latest ruby:alpine image contains bundler 2.2 series.
> So, this version of bundler starts to use `x86_64-linux-musl` platform instead of `ruby`.
> 
> Thus, we should fixed image version which is `ruby:alpine3.10` for generating Gemfile.lock, I guess.
> WDYT, @repeatedly?

I'm happy to make that change too, let me know.

repeatedly:
> Thus, we should fixed image version which is ruby:alpine3.10 for generating Gemfile.lock, I guess.

elsesiy:
@repeatedly @cosmo0920 I updated the PR, please have another look

cosmo0920:
Thanks for updating the PR.
I'd found another solution for ruby platform glitch:

```make
	docker run --rm -i -v $(PWD)/docker-image/$(DOCKERFILE)/Gemfile:/Gemfile:ro \
		ruby:alpine sh -c "apk add --no-cache --quiet git && bundle lock --print --remove-platform x86_64-linux-musl --add-platform ruby" > docker-image/${DOCKERFILE}/Gemfile.lock
```

Using `--remove-platform x86_64-linux-musl --add-platform ruby` arguments is better to handle newer bundler and ruby.

cosmo0920:
I'll push the commit which contains the above commented modification.

Issue #541, opened 2021-03-05
Newest version of the container does not start up due to config errors


WolfgangMehner:
We are running a fluentd Daemonset on our Kubernetes cluster (v.1.19.6 on EKS)

We started up the most recent version of the image yesterday, and everything worked fine. We used the tag with the resulting digest:
v1-debian-elasticsearch7 -> [18d5a7158b5b](https://hub.docker.com/layers/fluent/fluentd-kubernetes-daemonset/v1.12.0-debian-elasticsearch7-1.0/images/sha256-18d5a7158b5bcd2adc31f0968a769cd1cfcc4be2ab9432647eab94963373e44a?context=explore)
Today we started up containers with the same config and tag, but got a different container that did not start up anymore:
v1-debian-elasticsearch7 ->  [2a7e2d1e42f3](https://hub.docker.com/layers/fluent/fluentd-kubernetes-daemonset/v1-debian-elasticsearch7/images/sha256-2a7e2d1e42f32247447b6440c55b968a93c5501309aea52a458c2d1879443ca9?context=explore
)

When the container started up, we got an error along the lines of:
```
ConfigError error="<parse> section is required."
```
Unfortunately we can not provide the precise error message anymore, since nothing was logged with the fluentd PODs not starting up.

moorepatrick:
I'm seeing the same thing. A minor version was pushed last night but you were probably tagged on the major version. 
`v1-debian-elasticsearch7` is aliasing `v1.12.0-debian-elasticsearch7-1.1` today, so try `v1.12.0-debian-elasticsearch7-1.0`
(Note that you need the full `v1.12.0-debian`)

In my case I'm seeing the addition of `gem 'fluent-plugin-parser-cri' version '0.1.0'` at startup, which looks to be a likely culprit. Still digging into a real solution.

PR #542, opened 2021-03-05
Copy tail_container_parse.conf into Docker Images

moorepatrick:
Looking at the current Docker images built after https://github.com/fluent/fluentd-kubernetes-daemonset/pull/521 it looks like `tail_container_parse.conf` is not being copied over.

Fixes: https://github.com/fluent/fluentd-kubernetes-daemonset/issues/541


repeatedly:
Thanks!

PR #547, opened 2021-03-26
Add Fluentd v1.12.1 images

cosmo0920:


cosmo0920:
Before releasing Fluentd v1.12.2 images, we should release    Fluentd v1.12.1 images.

ashie (approved):

PR #558, opened 2021-04-13
change varlibdockercontainer path

rommelandrea:
In kubernetes 1.19+ the folder /var/lib/docker/ no more exist and the docker logs are placed in folder /var/logs/pods. 
This happen from the remove of dockershim and moby project layer

No changes on volumeMounts to avoid to edit all configs files

Tested on Azure Kubernetes Service 1.19.7 and 1.20.*

### DIFF
--- a/fluentd-daemonset-azureblob.yaml
+++ b/fluentd-daemonset-azureblob.yaml
@@ -104,4 +104,4 @@ spec:
           path: /var/log
       - name: varlibdockercontainers
         hostPath:
-          path: /var/lib/docker/containers
+          path: /var/log/pods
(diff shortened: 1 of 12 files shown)


fluentd-daemonset-azureblob.yaml
```diff
@@ -104,4 +104,4 @@ spec:
           path: /var/log
       - name: varlibdockercontainers
```
cosmo0920, line 105: We should change path names which are pointing out `/var/log/pods` instead of replacing paths.
These changes makes hard to debug and trace why `/var/log/pods` is mounted in `varlibdockercontainers`.

cosmo0920 (changes requested):

Issue #561, opened 2021-05-27
Follow new kubernetes logging mechanism


cosmo0920:
In kubernetes 1.19+ the folder `/var/lib/docker/` does not exist anymore and the docker logs are placed in folder `/var/logs/pods`.
This happens from the removal of `dockershim` and `moby` project layer.

This is originally reported in https://github.com/fluent/fluentd-kubernetes-daemonset/pull/558#issue-614415014.

Currently, fluentd kubernetes daemonset does not use containerd logging directories.
We should adopt new logging directories structure.

PR #562, opened 2021-05-28
Add Fluentd v1.12.4 images

ashie:
Signed-off-by: Takuro Ashie <ashie@clear-code.com>


cosmo0920 (approved):
LGTM. 

But I got CrashLoopBackoff when using v1.12.3 images on new kubernetes (1.19 or later).
The crash reason is:
https://github.com/fluent/fluentd-kubernetes-daemonset/issues/561
We should investigate how to fix it but we should merge this PR at first

PR #564, opened 2021-05-31
Updated kinesis plugin to latest version

tavisca-mgandhe:
Updated kinesis plugin to latest version so that the image can be used with the latest kinesis plugin which supports webTdentityTokenAuthentication , which is useful when using serviceAccount with Kubernetes


cosmo0920 (approved):
LGTM.
I didn't notice that kinesis plugin v3.4.0 is released. :eyes:

PR #566, opened 2021-06-08
unify multi-arch tags

### LINKED ISSUE #524: Unify multi-arch tags
Hi,

Please could you use manifests to provide a single tag name for all supported architectures? This allows our mixed-arch k8s cluster to automatically pull the correct image for each node.

I notice some previous bits relating to this have been closed, but Docker Hub _does_ support buildx manifests - they show it being used in their own blog post: https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

Capitrium:
This PR extends the automated DockerHub builds to produce a multi-arch image / manifest list in addition to the base per-architecture images by using [manifest-tool](https://github.com/estesp/manifest-tool) (based on [this example](https://github.com/ckulka/docker-multi-arch-example)) to create the manifest list instead of `docker manifest`, which is still unsupported in DockerHub builds.

These changes should be compatible with the existing DockerHub build settings, but note that both the amd64 and arm64 images must be available on DockerHub in order for `manifest-tool` to create the manifest list, so only the last architecture-specific build will be marked as `Successful` in the [build history](https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/builds).

Fixes #524.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,23 +12,23 @@
 
 IMAGE_NAME := fluent/fluentd-kubernetes
 X86_IMAGES := \
-	v1.12/debian-azureblob:v1.12.4-debian-azureblob-1.0,v1.12-debian-azureblob-1 \
-	v1.12/debian-elasticsearch7:v1.12.4-debian-elasticsearch7-1.0,v1.12-debian-elasticsearch7-1,v1-debian-elasticsearch \
-	v1.12/debian-elasticsearch6:v1.12.4-debian-elasticsearch6-1.0,v1.12-debian-elasticsearch6-1 \
-	v1.12/debian-loggly:v1.12.4-debian-loggly-1.0,v1.12-debian-loggly-1 \
-	v1.12/debian-logentries:v1.12.4-debian-logentries-1.0,v1.12-debian-logentries-1 \
-	v1.12/debian-cloudwatch:v1.12.4-debian-cloudwatch-1.3,v1.12-debian-cloudwatch-1 \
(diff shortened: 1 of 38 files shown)


cosmo0920:
I'm not familiar with multiarch image on DockerHub with manifest tool.
I understand that this PR can publish multiarch images but how do we set up building rules on DockerHub?
Current building rules are separated x86_64 and aarch64 images:
スクリーンショット 2021-06-09 10 53 51

Capitrium:
@cosmo0920 There aren't any new building rules that need to be set up on DockerHub using this method. It basically appends the `manifest-tool push` command to the end of both the existing x86 and aarch64 build post-push hooks; so `docker tag && docker push` will still run, but `manifest-tool push` will fail until both the x86 and aarch64 builds have been run (so the manifests that `manifest-tool` is referencing are pushed to DockerHub). This is also noted in the readme of the example repo [here](https://github.com/ckulka/docker-multi-arch-example/blob/master/README.md#automated-builds-via-docker-hub):

> No build rule for the (multi-arch) image tag in the manifest, e.g. latest
> ...
> Note that the builds may fail at the beginning until each build rule ran once. This can be because the manifest references
> images that don't exist yet. Hence, after running all of them, each referenced image exists and the manifest can be pushed
> succesfully.

I'm not actually familiar with manifest-tool either, or with builds on DockerHub, I'm just working off their documentation and the example in the [ckulka/docker-multi-arch-example](https://github.com/ckulka/docker-multi-arch-example) repo. 😆 😅

cosmo0920:
Hmm..., it's interesting.
Then I'll merge this PR and will revert if we got an issue by this change.

cosmo0920 (approved):

PR #570, opened 2021-06-16
Update kubernetes.conf

netgaori:


cosmo0920 (changes requested):
This kind of stuff should use ConfigMap to change/insert your custom configuration.
And also, tailing nginx log is not always used in Fluentd Daemonset images.
So, it might be unacceptable your change.

Anyway, thanks to consider contributing this repo. :muscle:

PR #575, opened 2021-07-02
Update kubernetes_metadata_filter to v2.7.1

kenhys:
Fluentd v1.13.0 or later is supported since v2.7.1


cosmo0920 (approved):
A big :+1:

cosmo0920:
@kenhys Then, you can push directly when bump up minor docker image version and execute `make src-all`.

PR #578, opened 2021-07-05
Explain "Configure Automated Builds" for every minor release

kenhys:
Not to trigger useless runners every minor release.


cosmo0920 (approved):
:+1:

PR #600, opened 2021-07-13
Configure dependabot behavior

kenhys:
It is introduced not to watch deprecated images.

ref. https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  # Watch dependencies for current stable
+  - package-ecosystem: "bundler"
+    directory: "/docker-image"
+    schedule:
+      interval: "daily"


.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+  # Watch dependencies for current stable
+  - package-ecosystem: "bundler"
+    directory: "/docker-image/v1.13"
```
ashie, line 10: We might forget to upgrade this...

ashie: So I prefer black list instead of white list.
(Move obsolete versions to other directory?)

.github/dependabot.yml
```diff
@@ -0,0 +1,12 @@
+version: 2
+updates:
+  # Maintain dependencies for GitHub Actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
```
cosmo0920, line 7: I think that daily interval is very bother for us.
weekly interval is better.
In other words, we wouldn't release this deamonset in weekly or daily, right?

ashie: I don't think so, actual vulnerability reports won't be so often.
And when a vulnerability is found, we should recognize it ASAP.

cosmo0920: Hmm, it's reasonable. Let's keep as-is.

.github/dependabot.yml
kenhys: Ah, moving obsolete versions is better, I mean that set `directory: "/docker-image/"` and move docker-image/v0.12 to somewhere.

.github/dependabot.yml
kenhys: Yes, "daily" is intended as above.

kenhys:
Moving obsolete recipes should be done in another PR.

ashie (approved):

cosmo0920 (approved):

PR #603, opened 2021-07-15
Move deprecated recipes under archived-image/

kenhys:
It is introduced to avoid false vulnerability detection by dependabot
because dependabot scans only docker-images/*.

It keeps stable(v1.13), old-stable(v1.12), old-old-stable (v1.11).


cosmo0920 (changes requested):
We also should add notification for deprecated images and Dockerfile locations.
At least, deprecation notification of old stable section should be handle in this PR: https://github.com/fluent/fluentd-kubernetes-daemonset#old-stable

kenhys:
Thanks, I've overlooked it.
Fixed.

cosmo0920 (approved):

PR #605, opened 2021-07-15
Use fixed memory leak issue of fluent-plugin-systemd

kenhys:
See
https://github.com/fluent-plugin-systemd/fluent-plugin-systemd/pull/91

TODO: make src-all when it is ready for newer release


cosmo0920 (approved):
Yup. We should use the newer version of systemd plugin. :runner:

PR #616, opened 2021-07-16
Specify explicit path which contains Gemfile

kenhys:
It seems that dependabot does not check recursively by `directory: "/docker-image"`,
so we need to define each rule for images.


cosmo0920 (approved):
Looks nice. :muscle:

PR #1081, opened 2021-07-17
Reduce watch version to latest one only

kenhys:
There is a case that pile of security alerts are raised.
This PR is aimed to reduce it.


kenhys:
Added open-pull-request-limit

templates/dependabot.yml.erb
```diff
@@ -8,11 +8,13 @@ updates:
     schedule:
       interval: "daily"
   # Watch dependencies for current stable only
-<% available_gemfile_paths = Dir.glob("docker-image/**/*/Gemfile") %>
+<% latest_version = Dir.glob("docker-image/*").sort.last %>
```
ashie, line 11: `sort` is inappropriate, `v1.10` will be placed former than `v1.9`.

kenhys: Ah, it should be fixed.

ashie (approved):

PR #1278, opened 2021-07-19
dependabot: Try to receive only security updates

ashie:
We don't intend to receive all updates, we want only security updates.
For this purpose, `open-pull-requests-limit: 0` seems effective because it doesn't affect to security updates:

https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates#open-pull-requests-limit

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,618 +7,721 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
+    open-pull-requests-limit: 0
   # Watch dependencies for current stable only
 
-
-
-  - package-ecosystem: "bundler"
-    directory: "/docker-image/v1.13/debian-loggly"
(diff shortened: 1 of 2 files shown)


kenhys (commented):
Using `open-pull-requests-limit: 0` is good approach. :+1: 

But does `directory: "/docker-image" work as intended?
(I recall that it caused an error before)

ashie:
> (I recall that it caused an error before)

What kind of error was it?

ashie:
Ah, dependabot see only Gemfile just under the directory?

ashie:
Finding it recursive doesn't seem supported yet: https://github.com/dependabot/dependabot-core/issues/2178

ashie:
Done at #1081

PR #1284, opened 2021-07-29
Use latest bundler to fix CVE-2020-36327

kenhys:
https://bundler.io/blog/2021/02/15/a-more-secure-bundler-we-fixed-our-source-priorities.html
As noted, at least bundler 2.2.18 or later must be used.

  Fix dependency confusion issues with implicit dependencies
  https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#security-fixes-1


cosmo0920 (approved):
:+1:

PR #1285, opened 2021-07-30
Remove namespace setting from ClusterRole resources

FaHeymann:
ClusterRoles are not namespaced and this might be misleading I think.


kenhys:
Basically, it is ok, but could you remove cosmetic changes (remove trailing spaces) from this PR?
It should be done with another PR (or separate commit)

kenhys (approved):
LGTM.

kenhys:
Thanks! @FaHeymann

PR #1286, opened 2021-07-30
Feature/debian-s3 support parquet compress

ChihSeanHsu:
For support parquet compress.
1. upgrade fluent-s3-plugin to 1.6.0
2. install columnfy 0.1.0

### DIFF
--- a/templates/Dockerfile.erb
+++ b/templates/Dockerfile.erb
@@ -29,7 +29,7 @@ ENV GEM_HOME /fluentd/vendor/bundle/ruby/2.6.0
 ENV FLUENTD_DISABLE_BUNDLER_INJECTION 1
 
 COPY Gemfile* /fluentd/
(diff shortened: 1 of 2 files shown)


kenhys:
Thanks!

but we don't directly modify these files.

Could you follow
https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/MAINTAINING.md#updating ?

ChihSeanHsu:
Hi @kenhys ,
Okay, I see. I'll modify it.
Do I need to re-generate all the different plugins or just for S3?

kenhys:
Use `<% if target == "s3" %>...<% end %>` in Dockerfile.erb, then regenerating just for s3 is enough.

ChihSeanHsu:
Hi @kenhys ,
Do I need to update all the files in docker-image or just these two template files?

kenhys (approved):
It's ok modifying template files.

LGTM.

kenhys:
Thanks  @chancez

PR #585, opened 2021-07-07
Remove duplicate config in conf/kubernetes.conf

ashie:
Follow up for #534
Fix #583

Signed-off-by: Takuro Ashie <ashie@clear-code.com>

PR #1297, opened 2021-09-02
Generate tail_container_parse.conf for each images

cosmo0920:
tail_container_parse.conf should be also handled in Makefile rules to generate for each images.

Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -310,6 +310,11 @@ systemd.conf:
 systemd.conf-all:
 	make each-image TARGET=systemd.conf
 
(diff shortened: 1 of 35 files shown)


kenhys (approved):
LGTM

ashie:
@cosmo0920 
tail_container_parse.conf doesn't exist in the template directory.
It's removed at #585.
In addition, it's not included anymore since #585
Should we revive it?

ashie:
#1298

PR #1298, opened 2021-09-06
Include tail_container_parse.conf

ashie:
Including tail_container_parse.conf was removed at #585 unexpectedly
and contents of the file is merged into kubernetes/containers.conf.
But the file doesn't exist, the following usage described in README.md
doesn't work as expected.

```
You can use `cri` parser by overwriting `tail_container_parse.conf` via ConfigMap.
```

Signed-off-by: Takuro Ashie <ashie@clear-code.com>

### DIFF
--- a/templates/conf/kubernetes/containers.conf.erb
+++ b/templates/conf/kubernetes/containers.conf.erb
@@ -6,8 +6,5 @@
   tag "#{ENV['FLUENT_CONTAINER_TAIL_TAG'] || 'kubernetes.*'}"
   exclude_path "#{ENV['FLUENT_CONTAINER_TAIL_EXCLUDE_PATH'] || use_default}"
   read_from_head true
-  <parse>
-    @type "#{ENV['FLUENT_CONTAINER_TAIL_PARSER_TYPE'] || 'json'}"
-    time_format "#{ENV['FLUENT_CONTAINER_TAIL_PARSER_TIME_FORMAT'] || '%Y-%m-%dT%H:%M:%S.%NZ'}"
-  </parse>
+  @include ../tail_container_parse.conf
 </source>


ashie:
The template file itself is already revived at 23607c65620f69d3d526a94347a3ae96dec913e9

templates/conf/kubernetes/containers.conf.erb
```diff
@@ -7,7 +7,6 @@
   exclude_path "#{ENV['FLUENT_CONTAINER_TAIL_EXCLUDE_PATH'] || use_default}"
   read_from_head true
   <parse>
-    @type "#{ENV['FLUENT_CONTAINER_TAIL_PARSER_TYPE'] || 'json'}"
-    time_format "#{ENV['FLUENT_CONTAINER_TAIL_PARSER_TIME_FORMAT'] || '%Y-%m-%dT%H:%M:%S.%NZ'}"
+    @include ../tail_container_parse.conf
   </parse>
```
cosmo0920, line 11: tail_container_parse.conf already uses `<parse>` directive. We should just include it here:

```
@include ../tail_container_parse.conf
```

cosmo0920 (changes requested):

templates/conf/kubernetes/containers.conf.erb
ashie: Sorry, I failed to amend the commit before pushing it.
I've fixed it.

cosmo0920 (approved):

cosmo0920:
I applied this patch at https://github.com/fluent/fluentd-kubernetes-daemonset/commit/555dba741e6c943983c19d7a4b618e10092b83ef and kicked k8s daemonset building on DockerHub.

PR #1312, opened 2021-10-19
Update container.conf to use env for path

MayankGandhe:
Update path to use env variable so that user have more flexibility and options in using the application and can also follow all the features of path declaration of fluentd


kenhys (approved):
LGTM.

ashie:
Thanks!

Issue #1328, opened 2022-01-08
Adding required assigments for FLUENT_UID=0 


sipirsipirmin:
`forwarder.containerSecurityContext.privileged=true` and  `forwarder.containerSecurityContext.allowPrivilegeEscalation=true`  is requiring  when FLUENT_UID=0 forwarders added to extraEnv. 

I think, this have to be mentioned in README

PR #1329, opened 2022-01-08
#1328 Run as root guide updated

sipirsipirmin:


cosmo0920:
Bitnami’s helm chart provides these options: https://github.com/bitnami/charts/blob/master/bitnami/fluentd/README.md#fluentd-parameters

But, our k8s daemonset does not provide them.
Could you explain how they work on our k8s daemonset?

PR #1342, opened 2022-02-18
Add arm64 version of fluent-plugin-opensearch image

cosmo0920:
We should also provide arm64 version of fluent-plugin-opensearch image.

Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>


ashie (approved):

cosmo0920:
BTW, we can drop the building rule of elasticsearch6 image? 
Elasticsearch 6 series is already reached EOL:
https://endoflife.date/elasticsearch

PR #1359, opened 2022-04-28
added OpenSearch template

marcio-pessoa:
I just created an OpenSearch template after suffering a bit trying to use the Elasticsearch template with OpenSearch.

Hope it can be helpful.


cosmo0920 (approved):
Looks good. Thanks for your work!

PR #1364, opened 2022-06-08
Add 'control-plane' key to DaemonSet toleration.

inerplat:
After upgrading my kubernetes cluster to v1.24, I confirmed that the fluentd daemonset is not assigned to the control-plane node.

This is due to the [change of Kubeadm](https://kubernetes.io/blog/2022/04/07/upcoming-changes-in-kubernetes-1-24/#api-removals-deprecations-and-other-changes-for-kubernetes-1-24), because Kubeadm changed the taint key from 'master' to 'control-plane'.

> [The master label is no longer present on kubeadm control plane nodes](https://github.com/kubernetes/kubernetes/pull/107533). For new clusters, the label 'node-role.kubernetes.io/master' will no longer be added to control plane nodes, only the label 'node-role.kubernetes.io/control-plane' will be added. For more information, refer to [KEP-2067: Rename the kubeadm "master" label and taint](https://github.com/kubernetes/enhancements/tree/master/keps/sig-cluster-lifecycle/kubeadm/2067-rename-master-label-taint). 

Therefore, I think that the toleration corresponding to the 'control-plane' key must be added to the fluentd-daemonset to operate in the same way as the original intention in clusters of v1.24 or higher version.

Even for clusters not configured with Kubeadm, I edited it to be added to all distribution types for migration compatibility.


ashie:
Thanks for your contribution!
Could you address failed check?: https://github.com/fluent/fluentd-kubernetes-daemonset/pull/1364/checks?check_run_id=6788191699
(Add `Signed-off-by:` to your commit by `git commit --amend -s`, then `git push -f`)

cosmo0920 (approved):
Looks reasonable for me.

ashie:
Thanks!

PR #1377, opened 2022-08-01
Feed fluent-kubernetes-metadata-plugin with the node name

orgads:
This reduces cache misses and needless calls to the Kubernetes API.

https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter#environment-variables-for-kubernetes


orgads:
ping?

ashie:
Thanks!

PR #1378, opened 2022-08-03
Avoid using async v2

ashie:
The base fluent/fluentd docker image for v1.15.1-1.0 include async v2 gem but fluentd isn't ready for it. To avoid using it, rebuild images with fluent/flunetd:v1.15.1-1.1.


ashie:
Waiting to compete rebuilding fluent/fluentd images.

PR #1386, opened 2022-09-08
Set flush_interval to 1 for forward and syslog images

orgads:
The default is 60, which results in a serious delay. Most of the other backends set a lower interval. Since the work of forward and syslog is minimal, flushing every second should be safe.

Signed-off-by: Orgad Shaneh <orgad.shaneh@audiocodes.com>


orgads:
ping

orgads:
ping

templates/conf/fluent.conf.erb
```diff
@@ -373,6 +373,7 @@
   packet_size 65535
 
   <buffer kubernetes_host>
+    flush_interval 1
```
cosmo0920, line 376: Could you make an environment variable pattern like as:

```
flush_interval "#{ENV['FLUENT_KUBERNETES_HOST_INTERVAL'] || use_default}"
```

To preserve the default behavior should be needed here.
Thanks for your understanding.

cosmo0920 (changes requested):

templates/conf/fluent.conf.erb
orgads: The default interval is 60s, and flushing to syslog is cheap and straight-forward. Why should we have 60s by default? Notice that other backends set much lower value:

| Name | Interval | Configurable by env var |
| ---- | -------- | ----------------------- |
| Elasticsearch | 5s | ✔ |
| OpenSearch | 60s | ✔ |
| StackDriver | 5s | ❌ |
| Gelf | 5s | ❌ |
| Logz.io | 5s | ❌ |
| Splunk | 10s | ✔ |
| Kinesis | 1s | ❌ |

cosmo0920: The main reason is backward compatibility for configurations.
Your change is obviously breaks for the default behavior.
And the minor reason is increasing to resend cost when using the small intervals.
Fluentd sending system should be robust by degin that is described into official doc:
https://docs.fluentd.org/buffer#the-lifecycle-of-chunks
When the writing chunks are failed to send to the some of storage systems, the backoff of chunks are frequently triggered. This is not intended behavior for some of users' log system.

orgads: I don't think there's any backoff/resend mechanism for syslog backend. Syslog is UDP, so there's no way to determine if it succeeded or failed. I couldn't find anything related to failure detection in https://github.com/fluent-plugins-nursery/fluent-plugin-remote_syslog

orgads: Ok, Added ENV variable, and changed the default to 5s. Is this acceptable?

ashie: Thanks.
IMHO It's acceptable, but not sure for other users.
Since they are now configurable, and we don't receive such claim from other users yet, I think we shouldn't change the default value for safety.
At least we shouldn't change the default value on tiny update even if we decide to change it (shouldn't apply for 1.16.x, should wait 1.17.x).

orgads: Ok, I changed the default to `use_default`. Resolving.

cosmo0920 (approved):

ashie (approved):

ashie:
Thanks!

PR #1392, opened 2022-11-02
Add v1.15.3 images

ashie:
If will also fix ##1391 by updating base debian image.


ondrejhlavacek:
Will this also fix the recent OpenSSL CVEs?

ashie:
> Will this also fix the recent OpenSSL CVEs?

Affected OpenSSL is v3: https://mta.openssl.org/pipermail/openssl-announce/2022-November/000243.html
We are using OpenSSL v1.1.1, so that it's not affected.

ondrejhlavacek:
@ashie Sure, my bad, sorry. Got mistaken by openssl 3.0.0 gem.

PR #1395, opened 2022-11-24
Add InsightOps support

Kavuti:
Since InsightOps is the successor of Logentries, I took an open source plugin for InsightOps and slightly modified to be similar to the Logentries' one.


Kavuti:
Hi, are there any news on this? I'm using it successfully in production, so it may be useful to have the official image supporting it.

Kavuti:
Hi, are there any news on this? I'm using it successfully in production, so it may be useful to have the official image supporting it.

Issue #1397, opened 2022-12-09
missing required field "selector" in fluentd-daemonset-papertrail


palminha:
```
carlos111@master:~/papertrail$ kubectl create -f fluentd-daemonset-papertrail.yaml 
error: error validating "fluentd-daemonset-papertrail.yaml": error validating data: ValidationError(DaemonSet.spec): missing required field "selector" in io.k8s.api.apps.v1.DaemonSetSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

running kubectl version 1.23.12
```
carlos111@master:~/papertrail$ kubectl version
Client Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.12", GitCommit:"c6939792865ef0f70f92006081690d77411c8ed5", GitTreeState:"clean", BuildDate:"2022-09-21T12:20:29Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"23", GitVersion:"v1.23.12", GitCommit:"c6939792865ef0f70f92006081690d77411c8ed5", GitTreeState:"clean", BuildDate:"2022-09-21T12:13:07Z", GoVersion:"go1.17.13", Compiler:"gc", Platform:"linux/amd64"}
carlos111@master:~/papertrail$ 
```

PR #1398, opened 2022-12-09
Fix missing selector in papertrail daemonset

palminha:
Fix issue #1397


ashie:
Sorry for my late response.
Thanks for your contribution!

PR #1410, opened 2023-01-30
Add pos_file prefix support

alantang888:
Let you able to run multiple fluentd on same node.

### DIFF
--- a/templates/conf/kubernetes/cluster-autoscaler.conf.erb
+++ b/templates/conf/kubernetes/cluster-autoscaler.conf.erb
@@ -3,7 +3,7 @@
   @id in_tail_cluster_autoscaler
   multiline_flush_interval 5s
   path /var/log/cluster-autoscaler.log
-  pos_file /var/log/fluentd-cluster-autoscaler.log.pos
+  pos_file "#{File.join('/var/log/', ENV.fetch('FLUENT_POS_EXTRA_DIR', ''), 'fluentd-cluster-autoscaler.log.pos')}"
   tag cluster-autoscaler
   <parse>
     @type kubernetes
--- a/templates/README.md.erb
+++ b/templates/README.md.erb
@@ -228,6 +228,12 @@ oc delete pod -l k8s-app=fluentd-logging
 
 This is from [nekop's japanese article](https://nekop.hatenablog.com/entry/2018/04/20/170257).
 
+### Multiple fluentd
+
+When you want to have multiple fluentd. For example push to multiple destination like: elsticsearch + S3.
+You need to use `FLUENT_POS_EXTRA_DIR` add additional directory for pos file.
+Otherwise they share same pos file. You may found some log only on one destination.
+
 ## Note
 
 ### kafka image doesn't support zookeeper parameters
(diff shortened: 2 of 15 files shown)


templates/conf/kubernetes/cluster-autoscaler.conf.erb
```diff
@@ -3,9 +3,9 @@
   @id in_tail_cluster_autoscaler
   multiline_flush_interval 5s
   path /var/log/cluster-autoscaler.log
-  pos_file /var/log/fluentd-cluster-autoscaler.log.pos
+  pos_file "#{'/var/log/' + ENV.fetch('FLUENT_POS_PREFIX', '') + 'fluentd-cluster-autoscaler.log.pos'}"
```
cosmo0920, line 6: How about using File.join to simplify to represent what you wanted to do like as?

```
pos_file "#{File.join('/var/log/', ENV.fetch('FLUENT_POS_PREFIX', ''), 'fluentd-cluster-autoscaler.log.pos')}"
```

Just joining as String instances is dangerous. Safely joining method should be needed to use here.

cosmo0920 (changes requested):

templates/conf/kubernetes/cluster-autoscaler.conf.erb
alantang888: I just tested a bit. Using File.join will create an addition directory. Not sure is that intention to have addition directory.
If that is the intention I can change to File.join.

---
And I just tested with `../..` with File.join. Look like it work same as string concat on this case. 

I set `FLUENT_POS_PREFIX` == `PREFIX` and `DFLUENT_POS_PREFIX` == `../../PREFIX`.

```
irb(main):001:0> File.join('/var/log/', ENV.fetch('FLUENT_POS_PREFIX', ''), 'fluentd-cluster-autoscaler.log.pos')
=> "/var/log/PREFIX/fluentd-cluster-autoscaler.log.pos"
irb(main):002:0> File.join('/var/log/', ENV.fetch('FLUENT_POS_PREFIX_NOT_EXIST', ''), 'fluentd-cluster-autoscaler.log.pos')
=> "/var/log/fluentd-cluster-autoscaler.log.pos"
irb(main):003:0> File.join('/var/log/', ENV.fetch('DFLUENT_POS_PREFIX', ''), 'fluentd-cluster-autoscaler.log.pos')
=> "/var/log/../../PREFIX/fluentd-cluster-autoscaler.log.pos"
```

alantang888: One more question. If added directory. Will fluentd create that directory?? Or will have exception??

cosmo0920: From the current Fluentd implementation, Fluentd will create directory if it doesn't exist:
https://github.com/fluent/fluentd/blob/master/lib/fluent/plugin/in_tail.rb#L239
This change will be safe.

alantang888: That's great. I modified to use File.join. And I changed the env name to `FLUENT_POS_EXTRA_DIR`. That should more fit for this behaviour.

cosmo0920 (approved):

cosmo0920:
Could you all take a look thisPR, @ashie @kenhys?
If you all are OK, merge and publish the minor bumping version images, please?

ashie:
Thanks!

PR #1413, opened 2023-02-14
Bump version to v1.15.3-1.1 to update base image

ashie:
We need to update base fluentd docker image before releasing this: https://github.com/fluent/fluentd-docker-image/pull/352


kenhys (approved):
LGTM.

ashie:
Thanks!

Issue #553, opened 2021-04-01
Remove fluent-plugin-google-cloud


ozayss:
Hello!

fluent-plugin-google-cloud is outdated in the debian-stackdriver docker image.

Google Cloud plugin 0.4.10. 
That's a version from 2015.
It does not work anymore.

Thank you!

cosmo0920:
Hi, thank you for your feedback.
I'm not familiar with Google Cloud Platform.

Which fluent-plugin-google-cloud plugin version did work well on GCP?
Did you try it another plugin version as custom built images?

ozayss:
I tried to build custom image  within fluent-plugin-google-cloud 0.10.7 (latest).
It requires fluentd version 1.11.2.

I tried this configuration and looks like it works fine.

cosmo0920:
Thanks for your reply.
We're waiting feedback and progress to be merged this PR:
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/447

Yeah, current issue is: fluent-plugin-google-cloud 0.10.7 refuses to work with Fluentd v1.12.

cosmo0920:
The new version of fluent-plugin-google-cloud will be released soon:
https://github.com/GoogleCloudPlatform/fluent-plugin-google-cloud/pull/447#issuecomment-827923571

We will publish this new fluent-plugin-google-cloud bundled image.

ashie:
I'm considering to remove fluent-plugin-google-cloud and debian-stackdriver image completely from this repository.
Recent fluent-plugin-google-cloud sticks required fluentd version to a exactly specified version, while this repository always follow latest fluentd version. So that we never able to sync these versions.

PR #1446, opened 2023-07-19
v1.16.2

ashie:


cosmo0920 (changes requested):
We need to get rid of elasticsearch6 image that is already EOL:
https://www.elastic.co/jp/support/eol
Note that 2023/08/01 or ES v9.0.0 released date should be EOL for Elasticsearch7.

ashie:
> We need to get rid of elasticsearch6 image that is already EOL: https://www.elastic.co/jp/support/eol Note that 2023/08/01 or ES v9.0.0 released date should be EOL for Elasticsearch7.

Thanks for notifying it.
Should it be done in this PR? 
I want to remove obsolete images ( like as #553) in another PR.
Elasicsearch6 image isn't pushed to Docker Hub since some time ago already.

cosmo0920 (approved):

ashie:
Thanks for your review!

Issue #434, opened 2020-04-30
[in_tail_container_logs] pattern not matched - tried everything, not sure what I am missing


kjroger94:
Hi, 

I am using https://github.com/fluent/fluentd-kubernetes-daemonset/tree/master/docker-image/v1.9/debian-cloudwatch as reference to install this image as a sidecar in a kubernetes pod. 

Added these additional ENVs
```
ENV FLUENTD_SYSTEMD_CONF disable
ENV FLUENTD_PROMETHEUS_CONF disable
ENV FLUENT_ELASTICSEARCH_SED_DISABLE "true"
```
along with Log group name and region for AWS Cloudwatch

_fluent.conf_
```
@include "#{ENV['FLUENTD_SYSTEMD_CONF'] || 'systemd'}.conf"
@include "#{ENV['FLUENTD_PROMETHEUS_CONF'] || 'prometheus'}.conf"
@include kubernetes.conf
@include conf.d/*.conf

<match **>
  @type cloudwatch_logs
  @id out_cloudwatch_logs
  log_group_name "#{ENV['LOG_GROUP_NAME']}"
  auto_create_stream true
  use_tag_as_stream true
  retention_in_days "7"
  json_handler yajl # To avoid UndefinedConversionError
  log_rejected_request "true" # Log rejected request for missing parts
</match>

```

_kubernetes.conf_

<label @FLUENT_LOG>
  <match fluent.**>
    @type null
  </match>
</label>

```
<source>
  @type tail
  @id in_tail_container_logs
  path /root/.pm2/logs/*.log
  pos_file /root/.pm2/sys.log.pos
  tag kubernetes.*
  exclude_path "#{ENV['FLUENT_CONTAINER_TAIL_EXCLUDE_PATH'] || use_default}"
  read_from_head true
  <parse>
    @type json
    time_format %Y-%m-%dT%H:%M:%S.%NZ
  </parse>
</source>
```
Fluentd is tailing the logs but no matter what config i try - 
```
2020-04-30 13:08:40 +0000 [warn]: #0 [in_tail_container_logs] pattern not matched: "\e[0mGET / \e[32m200 \e[0m45.213 ms - 170\e[0m"
2020-04-30 13:08:41 +0000 [warn]: #0 [in_tail_container_logs] pattern not matched: "{"
2020-04-30 13:08:41 +0000 [warn]: #0 [in_tail_container_logs] pattern not matched: "  \"level\": \"info\","
[...]

PR #1450, opened 2023-08-03
fix elastic templates so that tail log parser works and daemonsets dont crash

jamesla:
This fix provides 2 updates to the elastic daemonsets.

1. Solves an issue where there was an infinite loop because fluent logs were being exported to elastic which would create fluent entries...  and daemonsets were tending to crash.

2. Solves an issue where tail parser was not working because it did not have a parser associated. I have added a parser to parse docker style log entries as that's what most people are using, however if you are using something else it would be quite clear for a user to see how to modify it.

This is documented here (as well as other various github issues which appears to trip a lot of people up). https://github.com/fluent/fluentd-kubernetes-daemonset/issues/434

### DIFF
--- a/fluentd-daemonset-elasticsearch-rbac.yaml
+++ b/fluentd-daemonset-elasticsearch-rbac.yaml
@@ -89,6 +89,12 @@ spec:
             value: "elastic"
           - name: FLUENT_ELASTICSEARCH_PASSWORD
             value: "changeme"
+          # Prevents circular logging issue
+          - name: FLUENT_CONTAINER_TAIL_EXCLUDE_PATH
+            value: /var/log/containers/fluent*
+          # Docker log parser
+          - name: FLUENT_CONTAINER_TAIL_PARSER_TYPE
+            value: /^(?<time>.+) (?<stream>stdout|stderr)( (?<logtag>.))? (?<log>.*)$/
         resources:
           limits:
             memory: 200Mi
--- a/fluentd-daemonset-elasticsearch.yaml
+++ b/fluentd-daemonset-elasticsearch.yaml
@@ -56,6 +56,12 @@ spec:
             value: "ThisIsASuperLongToken"
           - name: LOGZIO_LOGTYPE
             value: "kubernetes"
+          # Prevents circular logging issue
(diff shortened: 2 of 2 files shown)


jamesla:
bump

fluentd-daemonset-elasticsearch-rbac.yaml
```diff
@@ -89,6 +89,12 @@ spec:
             value: "elastic"
           - name: FLUENT_ELASTICSEARCH_PASSWORD
             value: "changeme"
+          # Prevents circular logging issue
+          - name: FLUENT_CONTAINER_TAIL_EXCLUDE_PATH
+            value: /var/log/containers/fluent*
+          # Docker log parser
+          - name: FLUENT_CONTAINER_TAIL_PARSER_TYPE
+            value: /^(?<time>.+) (?<stream>stdout|stderr)( (?<logtag>.))? (?<log>.*)$/
```
ashie, line 97: It's not correct, it not valid for `@type` here:
https://github.com/fluent/fluentd-kubernetes-daemonset/blob/4066426d1a23da07e7f014ae0ca9693fdd1655a8/templates/conf/tail_container_parse.conf.erb#L2

It shoud be `cri` as described in https://github.com/fluent/fluentd-kubernetes-daemonset/issues/434#issuecomment-831801690

fluentd-daemonset-elasticsearch.yaml
```diff
@@ -56,6 +56,12 @@ spec:
             value: "ThisIsASuperLongToken"
           - name: LOGZIO_LOGTYPE
             value: "kubernetes"
+          # Prevents circular logging issue
+          - name: FLUENT_CONTAINER_TAIL_EXCLUDE_PATH
+            value: /var/log/containers/fluent*
+          # Docker log parser
+          - name: FLUENT_CONTAINER_TAIL_PARSER_TYPE
```
ashie, line 63: Ditto

ashie (changes requested):

ashie:
Should we make them default instead of replacing them in each configurations?

PR #1456, opened 2023-08-31
Update limitation about non-published s3 image

kenhys:


ashie:
s3 image should be published, it' a bug, not an intentional limitation.

https://hub.docker.com/repository/docker/fluent/fluentd-kubernetes-daemonset/tags?page=1&ordering=last_updated&name=s3

amd64 image for s3 is published but arm64 image isn't published.
This is why the tag v1.16-debian-s3-1 isn't created. We need both amd64 image and arm64 image to generate multi arch image.

ashie:
https://hub.docker.com/repository/docker/fluent/fluentd-kubernetes-daemonset/builds

s3 build settings don't exist.
We already reach the limitation of build config number, so some needed configuration seems dropped implicitly.
At least we need to revive s3 and syslog images, so we should drop some needless images.

kenhys:
Thanks, I've overlooked that situation.

PR #1476, opened 2024-01-17
Add v1.16.3-2.1 images

kenhys:
Usually in runtime use cases, there is no need to keep
development packages.

It causes hard to check false-positive vulnerability reports.

FYI: fluentd-docker-image already drops *-dev package.


ashie (approved):
LGTM

kenhys:
Thanks!

kenhys:
Oops, I've forgot continuous line, so I'll create another PR.

PR #1487, opened 2024-03-21
Add v1.16.4-1.0 images

kenhys:
Use latest v1.16.4 images

https://github.com/fluent/fluentd-docker-image/pull/377


ashie:
We'll skip v1.16.4 because it has a fatal race condition bug in processing buffer.

kenhys:
I'll retarget to 1.16.5 later.

PR #1491, opened 2024-03-29
Follow up fluentd v1.16.5-1.0 images

kenhys:
Since v1.16.5, new version is reset to amd64-1.0 or arm64-1.0,
so it should follow them.


ashie:
Oops, sorry.

ashie (approved):

PR #1495, opened 2024-06-11
Add v1.17.0-1.0 images

kenhys:
As fluent-plugin-kubernetes_metadata_filter has supported fluentd v1.17.0 since 3.5.0.


kenhys:
workflow should be modified after #1496 has been merged.

kenhys:
Updated target for v1.17 in workflow.
and re-generated dockerfiles.

daipom (commented):
Do we need these files?

* `docker-image/v1.17/{,arm64/}debian-elasticsearch6/`
* `docker-image/v1.17/{,arm64/}debian-stackdriver/`
* `plugins/v1.17/elasticsearch6/`
* `plugins/v1.17/stackdriver/`

Does the current Makefile make `docker-image/v1.17/debian-elasticsearch6/` and `docker-image/v1.17/debian-stackdriver/`?
Looks like they have been already removed from the Makefile.

daipom:
Sorry, I can't find how to manage `plugins/` in [MAINTAINING.md](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/b75789ce4e9bd53044177597a972cfdd440b6201/MAINTAINING.md) and don't understand it.
Do we need to manually create the corresponding empty directories?

kenhys:
plugins/v1.17 is treated as template and copy to docker-image/v1.17/ each images by plugins rule in Makefile.

daipom:
> Do we need these files?
> 
>     * `docker-image/v1.17/{,arm64/}debian-elasticsearch6/`
> 
>     * `docker-image/v1.17/{,arm64/}debian-stackdriver/`

The `arm64` files remain.
Shouldn't we remove those files as well?

kenhys:
Thanks, that's right.
I've remove it too.

daipom (approved):
LGTM. Thanks!

https://github.com/fluent/fluentd-kubernetes-daemonset/pull/1495#issuecomment-2188110208
> thus plugins has to be manually created every new v1.1x release.

Maybe we should add this flow to [MAINTAINING.md](https://github.com/fluent/fluentd-kubernetes-daemonset/blob/b75789ce4e9bd53044177597a972cfdd440b6201/MAINTAINING.md) later.

PR #1496, opened 2024-06-20
github: try to build and push for hub.docker.com

kenhys:
In the previous versions, we use "Docker Build Cloud". It is useful by kicking via Web UI to trigger pre-defined build pipeline.

Instead, there are limitation of pre-defined build pipelines. The max number of build pipelines is limited.

Thus, we can not maintain all of target images - some target or architecture such as syslog, s3/arm64 are missing.

And more, it means that we can't maintain v1.16 stable branch and v1.17 at the same time. Therefore we need to use alternative way - GitHub Actions.

To setup actions, set Repository secrets:

DOCKER_HUB_ORGS: fluent
DOCKER_HUB_USERNAME: someone's account
DOCKER_HUB_ACCESS_TOKEN: someone's access token

With using GitHub Actions, the following image can be published again:

* debian-elasticsearch7 arm64
* debian-logentries arm64
* debian-loggly arm64
* debian-logzio arm64
* debian-s3 arm64
* debian-papertrail 
* debian-papartrail arm64

main branch is aimed to build v1.17 for a while (#1495 )

PR #1497, opened 2024-06-24
github: backport docker build #1496

kenhys:
See PR#1496

Try to build and push v1.16.x images for hub.docker.com

In the previous versions, we use "Docker Build Cloud". It is useful by kicking via Web UI to trigger pre-defined build pipeline.

Instead, there are limitation of pre-defined build pipelines. The max number of build pipelines is limited to 25.

Thus, we can not maintain some target or architecture such as syslog, s3/arm64 and so on.

And more, it means that we can't maintain v1.16 stable branch and v1.17 at the same time. Therefore we need to use alternative way - GitHub Actions.

To setup actions, set Repository secrets:

DOCKER_HUB_ORGS: fluent
DOCKER_HUB_USERNAME: someone's account
DOCKER_HUB_ACCESS_TOKEN: someone's access token

With using GitHub Actions, the following image can be published again:

* debian-elasticsearch7 arm64
* debian-logentries arm64
* debian-loggly arm64
* debian-logzio arm64
* debian-s3 arm64
* debian-papertrail
* debian-papartrail arm64

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,205 @@
+name: Publish Docker Image
+
+on:
+  push:
+    branches:
+      - v1.16
+    tags:
+      - v1.*
+
+env:
+  REPOSITORY: ${{ secrets.DOCKER_HUB_ORGS }}/fluentd-kubernetes-daemonset
+jobs:
+  define-matrix:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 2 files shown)


kenhys:
synced with #1496

kenhys:
update with latest docker-build.yml.

daipom (approved):
LGTM. Thanks!

kenhys:
Now we can ready to ship specific target main(v1.17) and v1.16

PR #1500, opened 2024-06-25
github: use only tags

kenhys:
Backport PR#1499

Both of branches and tags are specified, either branches or tags are pushed, GitHub Actions are triggered.

In practical use-case, after some PRs has merged, we want to release it. Thus it is enough to specify tags.


daipom (approved):
LGTM.

Issue #1492, opened 2024-04-08
debian-kafka2 after v1.16.3-debian-kafka2-2.0 and above fails to start kafka output plugin


gpapaz:
### Describe the bug

All tags after the version of v1.16.3-debian-kafka2-2.0, i.e. v1.16.3-debian-kafka2-2.0, v1.16.3-debian-kafka2-2.1 and v1.16.5-debian-kafka2-1.0, suffer from this issue. They are unable to correctly start the rdkafka output plugin for fluentd.

In version **fluent/fluentd-kubernetes-daemonset:v1.16.3-debian-kafka2-2.0** when the container is started with the configuration mentione in the 'To reproduce' section the below errors can be seen:

```fluentd-1  | 2024-04-08 13:54:08 +0000 [warn]: #1 failed to flush the buffer. retry_times=3 next_retry_time=2024-04-08 13:54:16 +0000 chunk="615961fb1f6b351762de7dee1a9625a7" error_class=Rdkafka::Config::ConfigError error="Configuration property \"ssl.ca.location\" not supported in this build: OpenSSL not available at build time"
fluentd-1  |   2024-04-08 13:54:08 +0000 [warn]: #1 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:226:in `block (2 levels) in native_config'
fluentd-1  |   2024-04-08 13:54:08 +0000 [warn]: #1 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:216:in `each'
fluentd-1  |   2024-04-08 13:54:08 +0000 [warn]: #1 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:216:in `block in native_config'
fluentd-1  |   2024-04-08 13:54:08 +0000 [warn]: #1 <internal:kernel>:90:in `tap'
fluentd-1  |   2024-04-08 13:54:08 +0000 [warn]: #1 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:214:in `native_config'
[...]

PR #1503, opened 2024-07-10
Add v1.16.5-1.1 images

kenhys:
* debian-kafka, debian-kafka2: fix missing libsasl2-2 issue:

   Closes: #1492

   In the previous versions, "apt-get purge '*-dev'" is executed. It is intended to remove only needless development packages, but actually with --auto-remove option removes also runtime packages.
   This behavior is not intended at all, so drop --auto-remove option.


kenhys:
Revised again.

daipom (approved):
LGTM. Thanks!

kenhys:
Created v1.16.5-1.1 tag.

PR #1510, opened 2024-07-12
github: fix variable name

kenhys:
It should be github.ref_name insteaf of env.github.ref_name.


daipom (approved):
LGTM. Thanks. Sorry for my mistake.

PR #1518, opened 2024-07-18
github: add missing v1-debian-component and arm64 tag

kenhys:
In the previous versions, v1-debian-component and
v1-debian-component-arm64 tag were always created.

With GitHub Actions, it follows as before.

NOTE: v1-debian-component and v1-debian-component-arm64 should point latest version of fluentd. e.g. v1.17.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -172,6 +172,10 @@ jobs:
                 ;;
             esac
           done
+          # v1-debian-(component)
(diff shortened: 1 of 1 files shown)


kenhys:
Verified with forked one:

https://github.com/kenhys/fluentd-kubernetes-daemonset/actions/runs/9987812740/job/27603013691

image


https://hub.docker.com/repository/docker/kenhys/fluentd-kubernetes-daemonset/tags

image

daipom (approved):
LGTM! Thanks!

Issue #1520, opened 2024-07-31
debian-kafka2 still fails when using SSL parameters in configuration


gpapaz:
### Describe the bug

Continuing on #1492, although the rdkafka is now able to be loaded, when ssl is part of the configuration for connecting to the Kafka, it still fails with the below example error messages:

```
2024-07-31 15:06:24 +0000 [warn]: #0 Send exception occurred: Configuration property "ssl.ca.location" not supported in this build: OpenSSL not available at build time at /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:226:in `block (2 levels) in native_config'
2024-07-31T15:06:24.193641737Z 2024-07-31 15:06:24 +0000 [warn]: #0 failed to flush the buffer. retry_times=0 next_retry_time=2024-07-31 15:06:25 +0000 chunk="61e8c70098c0ee43f541c57c26bbe45f" error_class=Rdkafka::Config::ConfigError error="Configuration property \"ssl.ca.location\" not supported in this build: OpenSSL not available at build time"
2024-07-31T15:06:24.193837801Z   2024-07-31 15:06:24 +0000 [warn]: #0 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:226:in `block (2 levels) in native_config'
2024-07-31T15:06:24.193846071Z   2024-07-31 15:06:24 +0000 [warn]: #0 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:216:in `each'
2024-07-31T15:06:24.193862561Z   2024-07-31 15:06:24 +0000 [warn]: #0 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:216:in `block in native_config'
2024-07-31T15:06:24.193870861Z   2024-07-31 15:06:24 +0000 [warn]: #0 <internal:kernel>:90:in `tap'
2024-07-31T15:06:24.193877872Z   2024-07-31 15:06:24 +0000 [warn]: #0 /fluentd/vendor/bundle/ruby/3.2.0/gems/rdkafka-0.11.1/lib/rdkafka/config.rb:214:in `native_config'
[...]

PR #1521, opened 2024-07-31
#1520 Fix missing openssl deps (libssl-dev)

gpapaz:
Closes: #1520 
Adding the libssl-dev library in build dependencies allows the usage of SSL attributes in rdkafka2 configuration.


kenhys (approved):
Basically kafla/kafka2 should be updated, but it is okay.

kenhys:
Thanks @gpapaz !

PR #1522, opened 2024-08-13
Backport: fix missing OpenSSL dependency (PR#1521)

kenhys:
Closes: #1520

Adding the libssl-dev library in build dependencies allows the usage
of SSL attributes in rdkafka2 configuration.


daipom (approved):
LGTM. Thanks!

PR #1529, opened 2024-08-22
github: align matching component order

kenhys:
Before:

Even though tagged debian-kafka2, it matched *debian-kafka* pattern.

After:

Even though tagged debian-kafka2, it matched *debian-kafka2* correctly.


daipom (approved):
LGTM. Thanks!

PR #1530, opened 2024-08-22
Backport github: align match order

kenhys:
Before:

Even though tagged debian-kafka2, it matched *debian-kafka* pattern.

After:

Even though tagged debian-kafka2, it matched *debian-kafka2* correctly.


daipom (approved):
Oh... Sorry for my mistake...
Thanks for this fix!

kenhys:
It works as expected.

image

PR #1532, opened 2024-08-23
github: add workflow to check building images

kenhys:
It enables to build before tagging for releasing a newer image.
 Thus it does not push images by this workflow.

### DIFF
--- a/.github/workflows/check-build.yml
+++ b/.github/workflows/check-build.yml
@@ -1,0 +1,125 @@
+name: Check Docker Image Build
+
+on:
+  pull_request:
+
+env:
+  REPOSITORY: ${{ secrets.DOCKER_HUB_ORGS }}/fluentd-kubernetes-daemonset
+jobs:
+  define-matrix:
+    runs-on: ubuntu-latest
+    env:
+      ALL_COMPONENTS: >
+        [
+        "debian-azureblob",
+        "debian-cloudwatch",
+        "debian-elasticsearch7",
+        "debian-elasticsearch8",
+        "debian-forward",
+        "debian-gcs",
+        "debian-graylog",
+        "debian-kafka2",
+        "debian-kafka",
+        "debian-kinesis",
+        "debian-logentries",
+        "debian-loggly",
+        "debian-logzio",
+        "debian-opensearch",
+        "debian-papertrail",
+        "debian-s3",
+        "debian-syslog"
+        ]
+    outputs:
+      components: >
+        ${{ env.ALL_COMPONENTS }}
+    steps:
+      - name: Show target components
+        run: echo ${{ env.ALL_COMPONENTS }}
+  amd64:
+    needs: define-matrix
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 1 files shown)


kenhys:
As it can't access secrets. it fails.

kenhys:
It will be kicked like this:

daipom:
Thanks for this improvement!
Is there any reason why we use another yml file for the check?
(instead of skipping the existing pushing job on some condition?)

kenhys:
> Is there any reason why we use another yml file for the check?

This workflow focus on every pull request events, but existing workflow focus on tags.
Then it aimed to check whether the new commit breaks build process. 

Surely they have common recipe, but I thought that it is better to split it out.

daipom:
> > Is there any reason why we use another yml file for the check?
> 
> This workflow focus on every pull request events, but existing workflow focus on tags. Then it aimed to check whether the new commit breaks build process.
> 
> Surely they have common recipe, but I thought that it is better to split it out.

I see!
I am concerned about managing duplicate logic, but we can handle it if that becomes a problem.

If we do this with the same file, it seems we can fix it as follows.

```diff
diff --git a/.github/workflows/docker-build.yml b/.github/workflows/docker-build.yml
index 645ec38e..971529cf 100644
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,6 +1,7 @@
 name: Publish Docker Image
 
 on:
+  pull_request:
   push:
     tags:
       - v1.*
@@ -39,6 +40,9 @@ jobs:
       - id: arrange-components
         run: |
           echo "for_all=true" >> "$GITHUB_OUTPUT"
+          if [ "${{ github.ref_type }}" != tag ]; then
+            exit 0
+          fi
           for component in ${{ join(fromJSON(env.ALL_COMPONENTS), ' ') }}; do
             if [[ ${{ github.ref_name }} == *"$component"* ]]; then
               echo "for_all=false" >> "$GITHUB_OUTPUT"
@@ -83,7 +87,7 @@ jobs:
         with:
           context: ${{ env.CONTEXT }}
           provenance: false
-          push: true
+          push: ${{ startsWith(github.ref, 'refs/tags/v1.') }}
           platforms: linux/amd64
           tags: ${{ env.AMD64TAGS }}
           # dare to use old mediatype (application/vnd.docker.distribution.manifest.v2+json)
@@ -127,7 +131,7 @@ jobs:
         with:
           context: ${{ env.CONTEXT }}
           provenance: false
-          push: true
+          push: ${{ startsWith(github.ref, 'refs/tags/v1.') }}
           platforms: linux/arm64
           cache-from: type=gha
           cache-to: type=gha,mode=max
@@ -135,6 +139,7 @@ jobs:
           # dare to use old mediatype (application/vnd.docker.distribution.manifest.v2+json)
           outputs: oci-mediatypes=false
   manifest:
+    if: ${{ startsWith(github.ref, 'refs/tags/v1.') }}
     needs: [define-matrix, amd64, arm64]
     runs-on: ubuntu-latest
     strategy:
```

daipom (approved):
LGTM.

Issue #1537, opened 2024-10-10
Using CRI parser with filter_rt_kube_logs causes empty messages


rezonant:
### Describe the bug

When you use the [CRI parser](https://github.com/fluent/fluent-plugin-parser-cri) (`@type cri`) along with the daemonset deployment, the `filter_rt_kube_logs` record transformer assumes that the log message will be in `${record['log']}`, but the CRI parser outputs `message` not `log`. 

This drove us nuts for quite awhile before realizing what was wrong, because Papertrail showed no logs even though everything was sent to it -- its because Papertrail omits log events that have a blank message.

We changed it to this:

```
    <filter kubernetes.**>
      @type record_transformer
      @id filter_rt_kube_logs
      enable_ruby true
      <record>
        hostname #{ENV['FLUENT_HOSTNAME']}-${record["kubernetes"]["namespace_name"]}-${record["kubernetes"]["pod_name"]}
        program ${record["kubernetes"]["container_name"]}
        severity info
        facility local0
        message ${record['log'] || record['message']}
      </record>
    </filter>
```

...which resolves the issue.

### To Reproduce

Deploy this image with `tail_container_parse.conf` set to

```
    <parse>
      @type cri
    </parse>
```

Messages will be blank in the output.

### Expected behavior

Messages should not be blank in the output.

### Your Environment

```markdown
- Tag of using fluentd-kubernetes-daemonset: 
    * docker.io/fluent/fluentd-kubernetes-daemonset:v1-debian-papertrail
    * Hash: sha256:8f4f8279dd5b245b68ebe69a35a485b9fc9982bd4fe503f964ab01c6898211b2
    * fluentd 1.17.1
```

### Your Configuration

```apache
fluentd.conf


    @include kubernetes.conf
    @include systemd.conf

    # Capture all unmatched tags
    <match **>
      @type papertrail
      papertrail_host "#{ENV['FLUENT_PAPERTRAIL_HOST']}"
      papertrail_port "#{ENV['FLUENT_PAPERTRAIL_PORT']}"
    </match>
```

tail_container_parse.conf
```
    <parse>
      @type cri
    </parse>
```
```


### Your Error Log

```shell
No errors visible, but rsyslog receivers such as papertrail skip blank messages, so no logs are received.
```

daipom:
Fixed in #1541.

> It seems that we can't get `program` and `hostname` correctly when using CRI parser for text format CRI logs.
[...]

PR #1541, opened 2024-10-28
papertrail: add CRI parser fallback

kenhys:
Closes: #1537

When parser is replaced to CRI parser in tail_container_parse.conf, parsed output stored into message not log.
As a result, blank message will be omitted.

### DIFF
--- a/templates/conf/kubernetes.conf.erb
+++ b/templates/conf/kubernetes.conf.erb
@@ -61,7 +61,7 @@
     program ${record["kubernetes"]["container_name"]}
     severity info
     facility local0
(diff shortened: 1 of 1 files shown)


kenhys:
TODO: create another PR to add images.

daipom (approved):
Thanks!

It seems that we can't get `program` and `hostname` correctly when using CRI parser for text format CRI logs.
It would be necessary to overwrite these settings according to the situation.

On the other hand, it would be fatal for `message` to be blank, so we should take care of `message` by the default setting.

PR #1543, opened 2024-10-28
Drop unused post_push from src: target

kenhys:
docker images were now build with GitHub Actions.

Previously post_push was used with Automated Builds on hub.docker.com. It was not required anymore.

For a record, post_push.erb was kept for a while, but marked as deprecated.


daipom (approved):
LGTM. Thanks!

PR #1544, opened 2024-11-22
Add OpenSearch sniffer class name and use it by default

slopezz:
The same done for Elasticsearch at https://github.com/fluent/fluentd-kubernetes-daemonset/pull/403, but now for OpenSearch.

Both Elasticsearch/Opensearch Sniffer class works the same way, need to be added to fluentd entrypoint https://github.com/fluent/fluent-plugin-opensearch?tab=readme-ov-file#tips.

It is required when fluentd don't connect directly to Opensearch servers, but connect to a proxy or load balancer in front of Opensearch servers.

Signed-off-by: Sergio López <slopezma@redhat.com>


slopezz:
Hi @kenhys , any chance to merge this?

ashie (approved):
LGTM

Sorry for overlooking this.
It should be shipped in next release.

kenhys (approved):
LGTM.

Issue #1546, opened 2024-12-10
Initialization of kubeclient fails when running in IPv6 single stack cluster


frauniki:
### Describe the bug

Initialization of kubeclient fails when running in IPv6 single stack cluster.

Here I am trying to fill `kubernetes_url` from ENV, but I get an error because the ipv6 address url needs to be escape by `[]`.
https://github.com/fluent/fluentd-kubernetes-daemonset/blob/master/templates/conf/kubernetes.conf.erb#L31

Change `kubernetes_url` to the following or remove it from the config and it will work correctly.
```
kubernetes_url "#{ENV['FLUENT_FILTER_KUBERNETES_URL'] || 'https://[' + ENV.fetch('KUBERNETES_SERVICE_HOST') + ']:' + ENV.fetch('KUBERNETES_SERVICE_PORT') + '/api'}"
```

### To Reproduce

Just deploy this container image to an IPv6 single stack kubernetes cluster and it should reproduce.

### Expected behavior

fluentd pod starts successfully.

### Your Environment

```markdown
- Tag of using fluentd-kubernetes-daemonset: fluent/fluentd-kubernetes-daemonset:v1.17.1-debian-s3-1.2
```

### Your Configuration

```apache
None (default config)
```

### Your Error Log

```shell
{"time":"2024-12-10 06:13:56 +0000","level":"info","message":"gem 'fluentd' version '1.16.6'"}
{"time":"2024-12-10 06:13:57 +0000","level":"info","message":"adding rewrite_tag_filter rule: message [#<Fluent::PluginHelper::RecordAccessor::Accessor:0x00007fcea304da48 @keys=\"message\">, /\\\"tag\\\":\\\"(\\w+)\\\"/, \"\", \"categorized.${tag_parts[1]}.${tag_parts[2]}.${tag_parts[3]}.${tag_parts[4]}.$1\", nil]"}
/usr/local/lib/ruby/3.2.0/uri/rfc3986_parser.rb:66:in `split': bad URI(is not URI?): "https://fdad:c5a9:94e1::1:443/api" (URI::InvalidURIError)
    from /usr/local/lib/ruby/3.2.0/uri/rfc3986_parser.rb:71:in `parse'
    from /usr/local/lib/ruby/3.2.0/uri/common.rb:193:in `parse'
[...]

PR #1548, opened 2024-12-18
Fix failure on IPv6 single stack cluster

kenhys:
When FLUENT_FILTER_KUBERNETES_URL is not set, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT are used as a fallback.

When applying such a fallback, it will fail on only IPv6 is available. It causes the following error:

  /usr/local/lib/ruby/3.2.0/uri/rfc3986_parser.rb:66:in `split': bad
  URI(is not URI?):
  "https://IPV6_ADDRESS:443/api" (URI::InvalidURIError)

IPv6 address must be enclosed with square brackets.

Closes: #1546

### DIFF
--- a/templates/conf/kubernetes.conf.erb
+++ b/templates/conf/kubernetes.conf.erb
@@ -28,7 +28,7 @@
 <filter kubernetes.**>
   @type kubernetes_metadata
   @id filter_kube_metadata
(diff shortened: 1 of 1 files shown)


ashie (approved):

kenhys:
Thanks!

PR #1549, opened 2024-12-18
ci: drop needless login action

kenhys:
It is needless to login action just to build for checking purpose.


kenhys:
ref.

https://github.com/kenhys/fluentd-kubernetes-daemonset/pull/887

kenhys:
This might be troublesome feedback from contributor.

daipom (approved):
LGTM. Thanks!

PR #1560, opened 2025-02-03
Backport(v1.16) papertrail: add CRI parser fallback (#1541)

kenhys:
Closes: #1537

When parser is replaced to CRI parser in tail_container_parse.conf, parsed output stored into message not log.
As a result, blank message will be omitted.


daipom (approved):
Thanks!

PR #1561, opened 2025-02-03
Backport(v1.16) Add OpenSearch sniffer class name and use it by default (#1544)

kenhys:
The same done for Elasticsearch at https://github.com/fluent/fluentd-kubernetes-daemonset/pull/403, but now for OpenSearch.

Both Elasticsearch/Opensearch Sniffer class works the same way, need to be added to fluentd entrypoint https://github.com/fluent/fluent-plugin-opensearch?tab=readme-ov-file#tips.

It is required when fluentd don't connect directly to Opensearch servers, but connect to a proxy or load balancer in front of Opensearch servers.

Signed-off-by: Sergio López <slopezma@redhat.com>


daipom (approved):
Thanks

PR #1562, opened 2025-02-03
Bump base arm64 image

kenhys:
Recently QEMU on amd64 host cause build failure, so
we switched to ubuntu 24.04 arm host.

This is because that -arm64-1.0 will not published.
Use -arm64-1.1 for k8s.

https://github.com/fluent/fluentd-docker-image/pull/414
https://github.com/fluent/fluentd-docker-image/pull/417

Let's try to ship v1.16.7-1.0 again.


kenhys:
Failed with previous try:

https://github.com/fluent/fluentd-kubernetes-daemonset/actions/runs/13105587386

daipom (approved):
Thanks!

PR #1564, opened 2025-02-10
Update target in README.md

kenhys:
As migrated from autobuilds on hub.docker.com,
There is no no-built images restrictions anymore.

It should be updated.


Watson1978 (approved):
👍🏻

daipom (approved):
Thanks!

PR #1558, opened 2025-01-31
github: use arm64 host directly to publish image

kenhys:

PR #1567, opened 2025-02-12
Add v1.18.0-1.2 images

kenhys:
In v1.18.0-1.2 release highlight:
 
* Add OpenSearch sniffer class name and use it by default
* Use arm64 host directly to publish image (#1558) 
* Unexpectedlly, v1.17 images was tagged as v1.18.0 series, so re-release them (#1566)
* Update target in README.md (#1564)


daipom (approved):
Thanks!

PR #1570, opened 2025-02-28
Add v1.16.7-1.1 images 

kenhys:
* Bump version to 1.16.7-1.1
* Add v1.16.7-1.1 images
* Reduce memory usage spike with fluent-plugin-systemd 1.1.1 (systemd-journal 2.1.0)


daipom:
I have noticed that we need to backport the following fixes (it could be separate from this PR).

* 412dca33ad421eb5f9a61cf4396ad49dc3f5dcca
* 8b5848269fe65c7057db507adc3427c93bc9aea3

daipom (approved):
Thanks!

PR #1571, opened 2025-03-13
Add v1.18-1.4 images

kenhys:
Fix some Out-of-bound Read in Ruby JSON Parser
(should be json 2.10.2 or later)

See https://github.com/advisories/GHSA-9m3q-rhmv-5q44


Watson1978 (approved):
👍🏻

PR #1579, opened 2025-06-19
Update fluent-plugin-record-modifier

BeeeeeMo:
Update fluent-plugin-record-modifier to 2.2.0


daipom (approved):
Thanks!
fluent-package is already using this version.
Let's update it as well.

PR #1624, opened 2026-04-08
github: tweak dependabot frequency

kenhys:
Before:

  * "Dependabot Security Alerts" is enabled
  * Security Alerts scans All contents, so it creates noisy PRs.

  After:

  * Disable "Dependabot Security Alerts".
  * Check daily and security updates PR at once.
  * Check daily and normal updates PR will be created in one PR. It will be rebased if it was let alone.


kenhys:
"Dependabot Security Alerts" was disabled.

.github/dependabot.yml
```diff
@@ -1,265 +1,47 @@
...
+      monthly-updates:
+        applies-to: version-updates
+        patterns:
+          - "*"
+    # monthly-updates will be grouped in one PR.
+    open-pull-requests-limit: 1
+
+  # security updates in daily, feature updates in monthly
+  - package-ecosystem: "bundler"
+    directory: "/docker-image/docker-image/v1.19"
```
Watson1978: Seems this has wrong directory path.

```suggestion
    directory: "/docker-image/v1.19"
```

it need to fix `templates/dependabot.yml.erb`

Watson1978 (approved):
👍🏻

PR #1671, opened 2026-06-30
Release v1.19.3 1.0

kenhys:


docker-image/v1.19/arm64/debian-azureblob/Gemfile
```diff
@@ -3,7 +3,7 @@
 
 source "https://rubygems.org"
 
-gem "fluentd", "1.19.2"
+gem "fluentd", "1.19.3"
 gem "oj", "~> 3.16.11"
```
Watson1978: I think it is better with updating oj.
Ref. https://github.com/fluent/fluentd-docker-image/pull/499

Watson1978 (approved):
👍🏻

PR #1676, opened 2026-07-23
feat: add v1.19 debian-elasticsearch9 image support

casabre:
## Summary
- Add missing Elasticsearch 9 support to the `v1.19` daemonset image matrix by introducing `debian-elasticsearch9` for both amd64 and arm64.
- Wire the new image into release/build surfaces (`Makefile`, Docker publish workflow, README, dependabot) while keeping the existing `v1-debian-elasticsearch` alias on Elasticsearch 7 for backward compatibility.
- Configure Fluentd output for ES9 with `use_elasticsearch_version 9` and remove `type_name` for the ES9 path.

## Why this is needed
Elasticsearch 9 support is currently missing from the repository's maintained `v1.19` image line. Teams moving from Elasticsearch 7/8 to 9 cannot use an official daemonset variant today, which forces custom downstream workarounds.

This PR adds an upstream ES9 image path with minimal behavioral change to existing variants.

## Dependency and compatibility notes
A locally validated working gem version set is:
- `fluent-plugin-elasticsearch 6.0.0`
- `elasticsearch 9.4.3`
- `elasticsearch-api 9.4.3`
- `elastic-transport 8.5.3`

With Fluentd output configured as:
- `use_elasticsearch_version 9`

The ES9 Gemfiles intentionally rely on transitive dependencies from `fluent-plugin-elasticsearch` 6.0.0 to keep the change set low profile.

## Validation
- Built successfully in WSL:
  - `docker build docker-image/v1.19/debian-elasticsearch9`
- Verified runtime dependency resolution in the built image:
  - `fluent-plugin-elasticsearch (6.0.0)`
  - `elasticsearch (9.4.3)`
- Verified container process starts with `FLUENTD_CONF=disable.conf`.


templates/conf/fluent.conf.erb
```diff
@@ -19,10 +19,13 @@
...
    scheme "#{ENV['FLUENT_ELASTICSEARCH_SCHEME'] || 'http'}"
    ssl_verify "#{ENV['FLUENT_ELASTICSEARCH_SSL_VERIFY'] || 'true'}"
-<% if target == "elasticsearch7" %>
+<% if target == "elasticsearch7" || target == "elasticsearch9" %>
    ssl_version "#{ENV['FLUENT_ELASTICSEARCH_SSL_VERSION'] || 'TLSv1_2'}"
 <% else %>
    ssl_version "#{ENV['FLUENT_ELASTICSEARCH_SSL_VERSION'] || 'TLSv1'}"
+<% end %>
+<% if target == "elasticsearch9" %>
+   use_elasticsearch_version 9
```
kenhys: Intended `default_elasticsearch_version`?

kenhys (changes requested):
Additionally missing plugins/v1.19/elasticsearch9
It causes `make src-all` failure.

templates/conf/fluent.conf.erb
casabre: Yes, actually this was intended because it was my working scenario. If it is not useful, I can remove it.

casabre:
> Additionally missing plugins/v1.19/elasticsearch9 It causes `make src-all` failure.

Will have a look. I passed for my local setup.

templates/conf/fluent.conf.erb
kenhys: appropriate parameter should be used here IMHO

casabre: Fixed in 276a648c:

1. Replaced `use_elasticsearch_version` with the correct plugin parameter `default_elasticsearch_version 9`.
2. Added `plugins/v1.19/elasticsearch9/.gitkeep` so `make src-all` / `make plugins` can copy assets for the new target.

casabre:
Addressed the review feedback in 276a648c:

- Config: switched to `default_elasticsearch_version 9` (the actual fluent-plugin-elasticsearch parameter).
- Plugins: added `plugins/v1.19/elasticsearch9/.gitkeep` to fix `make src-all` / plugins copy for the new target.

templates/entrypoint.sh.erb
```diff
@@ -70,7 +70,7 @@ else
             export FLUENT_CONTAINER_TAIL_EXCLUDE_PATH="$FLUENT_CONTAINER_TAIL_EXCLUDE_PATH,$IGNORE_TAIL_PATH"
         fi
     else
-        # No exclude rule yet, set ignore 
+        # No exclude rule yet, set ignore
```
kenhys: Good catch, but please not include this PR.
like these cosmetic change is recommended as separated PR.

docker-image/v1.19/arm64/debian-elasticsearch9/entrypoint.sh
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env sh
+
+# AUTOMATICALLY GENERATED
+# DO NOT EDIT THIS FILE DIRECTLY, USE /templates/entrypoint.sh.erb
```
kenhys, line 4: Expected permission of entrypoint.sh is 755, not 644

kenhys (changes requested):

templates/entrypoint.sh.erb
casabre: Reverted the trailing-whitespace cosmetic change in 7044f8aa. The only remaining entrypoint.sh.erb change in this PR is adding elasticsearch9 to the sniffer branch.

docker-image/v1.19/arm64/debian-elasticsearch9/entrypoint.sh
casabre: Fixed in 7044f8aa: set both amd64 and arm64 debian-elasticsearch9/entrypoint.sh to mode 755.

casabre:
Addressed the latest review feedback in 7044f8aa:

- Reverted the cosmetic trailing-whitespace change in templates/entrypoint.sh.erb (kept only the functional elasticsearch9 sniffer condition).
- Set entrypoint.sh permissions to 755 for both amd64 and arm64 ES9 images.

kenhys (approved):

kenhys:
Thank you for contributing!

kenhys:
Now elasticsearch9 images are available.
https://hub.docker.com/r/fluent/fluentd-kubernetes-daemonset/tags?name=elasticsearch9

PR #1681, opened 2026-08-03
Fix ignore pattern

kenhys:
Drop useless ignore and relax actions interval

update-types accepts version-update:semver-major,
     version-update:semver-minor, version-update:semver-patch, but
     none of them is specified. It seems that it breaks monthly-update for
     github actions. (guess)

ref. https://docs.github.com/code-security/reference/supply-chain-security/dependabot-options-reference#ignore


Watson1978 (approved):
👍🏻

PR #1682, opened 2026-08-04
chore: remove trailing whitespace in entrypoint comment

casabre:
## Summary
- Remove trailing whitespace on the `# No exclude rule yet, set ignore` comment in `templates/entrypoint.sh.erb`.
- Apply the same cosmetic cleanup to the generated `docker-image/v1.19/**/entrypoint.sh` files.

## Why a separate PR
Split out from review feedback on #1676 so the ES9 feature PR stays focused on functional changes.

## Test plan
- [ ] Confirm diff is comment-whitespace only (no behavioral changes)
- [ ] Spot-check one amd64 and one arm64 generated entrypoint against the template


kenhys (approved):

kenhys:
Thanks!

